PR #5, opened 2020-08-07
Plt 4613

amhudson:
Closes #

PLT-4613

#### Changelog

**New**

-Creating Flow Global Configuration Controller/Service
-Create FlowSettings/Entity/Service/Repository
-added mongo.encrypt.secret, mongo.encrypt.salt properties
-added and enabled security.platform.roles property

**Changed**

- None

**Removed**

- None

### DIFF
--- a/src/main/java/net/boomerangplatform/controller/ConfigurationController.java
+++ b/src/main/java/net/boomerangplatform/controller/ConfigurationController.java
@@ -12,7 +12,7 @@ import net.boomerangplatform.model.FlowSettings;
 import net.boomerangplatform.service.crud.ConfigurationService;
 
 @RestController
-@RequestMapping("/flow/settings/config")
+@RequestMapping("/flow/settings")
 @Secured({"ROLE_admin"})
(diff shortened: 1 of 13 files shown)


src/main/java/net/boomerangplatform/controller/ConfigurationController.java
```diff
@@ -0,0 +1,34 @@
...
+import org.springframework.web.bind.annotation.RequestBody;
+import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.RestController;
+import net.boomerangplatform.model.FlowSettings;
+import net.boomerangplatform.service.crud.ConfigurationService;
+
+
+
+@RestController
+@RequestMapping("/ci/settings/config")
```
timrbula, line 17: Can this just be `flow/settings` ? Is the /config necessary?

amhudson: updated

PR #27, opened 2020-10-30
Allowing configuration of mongo collection names with a prefixes

marcusdroy:
**New**

- Allows prefixing of of collection names through new spring boot property 'workflow.mongo.collection.prefix'

### DIFF
--- a/src/main/resources/application.properties
+++ b/src/main/resources/application.properties
@@ -117,3 +117,6 @@ workflow.service.host=localhost:${server.port}
 spring.aop.proxy-target-class=true
 
 logging.level.net.boomerangplatform.service.refactor=DEBUG
+
+workflow.mongo.collection.prefix=cicd
+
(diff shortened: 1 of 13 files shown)


src/main/resources/application.properties
```diff
@@ -117,3 +117,6 @@ workflow.service.host=localhost:${server.port}
 spring.aop.proxy-target-class=true
 
 logging.level.net.boomerangplatform.service.refactor=DEBUG
+
+workflow.mongo.collection.prefix=cicd
```
tlawrie, line 121: should this default to flow, and then we adjust to CICD?

tlawrie:
QQ for you. How does `#{@mongoConfiguration.prefix(` map to `workflow.mongo.collection.prefix`

marcusdroy:
> QQ for you. How does `#{@mongoConfiguration.prefix(` map to `workflow.mongo.collection.prefix`

MongoConfiguration component prefix method will get called to produce a new collection name with the prefix.

src/main/resources/application.properties
marcusdroy: I did intend this to be blank instead of cicd.

PR #96, opened 2021-08-03
Fix 91 tokenvalidationresponse

tlawrie:
**Changed**

- Return 403 if workflow is not found in validateToken


tlawrie:
Hi @marcusdroy what do we want to do with this PR?

PR #99, opened 2021-08-30
Feature/jetstream

corcoja:
Closes boomerang-io/roadmap#194
Integrates changes in PR boomerang-io/flow.lib.streaming#1

### Changelog

**New**

- NATS Jetstream integration for listening for incoming messages (CloudEvents) from a pull-based consumer.

**Changed**

- Readme.
- _Beautify_ application properties.

**Removed**

- Old NATS client.
- Duplicate Maven dependencies inside `pom.xml`.

### Testing / Reviewing

See testing for boomerang-io/flow.lib.streaming#1.


src/main/java/io/boomerang/client/JetstreamListener.java
```diff
@@ -0,0 +1,40 @@
...
+  private String jetstreamStreamSubject;
+
+  @Autowired
+  private EventProcessor eventProcessor;
+
+  @EventListener(ApplicationReadyEvent.class)
+  void onApplicationReadyEvent() throws InterruptedException {
+
+    if (jetstreamClient.subscribe(jetstreamStreamSubject, ConsumerType.PullBased,
+        (subject, message) -> eventProcessor.processNATSMessage(message))) {
```
tlawrie, line 32: Hey @corcoja two questions:

1. What drives the connection to remain stable? Do we need to implement the IdleHeartbeat?

2. What is the pull polling period? Is there a configuration setting for that?

corcoja: Hey @tlawrie, great questions. My idea for `flow.lib.streaming` library (PR boomerang-io/flow.lib.streaming#1), was to create a simple interface, easy to integrate, that ensures the code **just works**. All the hard work is done by the library.

## 1. What drives the connection to remain stable? Do we need to implement the IdleHeartbeat?

NATS `Connection` is configured here:
https://github.com/boomerang-io/flow.lib.streaming/blob/58870bfadb5d24ab2776e1bb84bd9b79976738d6/src/main/java/io/boomerang/jetstream/ConnectionAutoConfiguration.java#L57

From docs:

> The Java client will automatically reconnect if it loses its connection the nats-server. If given a single server, the client will keep trying that one. If given a list of servers, the client will rotate between them.

The options I included when configuring a NATS `Connections`, are:

- `eventing.jetstream.server.reconnect-wait-time`, defaults to `PT10S`.
- `eventing.jetstream.server.reconnect-max-attempts`, defaults to `-1` (no limit).

So, if NATS server goes down, the client will try to reconnect indefinitely each 10 seconds until the server comes back online. Of course, during this time, the client message consumer will receive no messages and no messages can be pushed to the stream, but the latter is backed up by the mechanism which forwards the message to `flow-controller` if NATS Jetstream fails.

Additionally, Java NATS Client supports Jetstream heartbeats (https://docs.nats.io/jetstream/concepts/consumers#idleheartbeat and https://github.com/nats-io/nats.java/blob/main/src/examples/java/io/nats/examples/jetstream/NatsJsPushSubHeartbeat.java) but this is related to push-consumers, which acknowledges them there are no new messages in the queue. From docs:

> If the idle heartbeat period is set, the server will send a status message with to the client when the period has elapsed but it has not received any new messages. This lets the client know that it's still there, but just isn't receiving messages.

We are using pull-based consumers.

## 2. What is the pull polling period? Is there a configuration setting for that?

Under the hood implementation uses a pull-based consumer in the form of a Java `Iterator`, wrapped inside an infinite `while` loop (unless an exception is thrown, like the subscription has been cancelled). The `Iterator` object is returned by blocking method with a timeout.

**How it works?**

Inside the `while` loop, the client will wait for a batch of `eventing.jetstream.consumer.pull.batch-size` (defaults to `10`) messages for at most `eventing.jetstream.consumer.pull.batch-first-message-wait` time (defaults to `PT30S`) from the server:

- If during these 30 seconds the NATS server receives at least on message, this message will be pulled by the client.
- If the NATS server has more than the default `10` messages, these awaiting messages will be pulled by the client, processed by the client and then immediately a new loop starts, repeating the process until no messages are on the NATS server's queue.
- If during the configured 30 seconds no new messages are pushed to the NATS server, i. e. client has nothing to pull, the method will time-out (unblock the execution stack), but the `while` loop will continue, requesting new messages again, indefinitely.

- Pull-based Java `Iterator` example: https://github.com/nats-io/nats.java/blob/main/src/examples/java/io/nats/examples/jetstream/NatsJsPullSubIterate.java
- `flow.lib.stream` code for pull-based consumers: https://github.com/boomerang-io/flow.lib.streaming/blob/58870bfadb5d24ab2776e1bb84bd9b79976738d6/src/main/java/io/boomerang/jetstream/JetstreamClientImpl.java#L180

tlawrie: Add to the README of the Library and also a link to that in the Workflow README

PR #132, opened 2021-12-13
Feat 241 multischeduletriggers 3.5

tlawrie:
Relates to boomerang-io/roadmap#241

Implements new WorkflowSchedules to allow Workflows to have multiple difference future run once and recurring schedule triggers.

_Note:_ this is the [prior PR](https://github.com/boomerang-io/flow.service.workflow/pull/124) based off of Release 3.5

#### Changelog

**New**

- {{new thing}}

**Changed**

- {{change thing}}

**Removed**

- {{removed thing}}

#### Testing / Reviewing

{{ Add descriptions, steps or a checklist for how reviewers can verify this PR works or not }}

PR #126, opened 2021-11-05
Feature/lib.eventing v0.2

corcoja:
Closes boomerang-io/flow.service.listener#15 

This PR upgrades eventing integration for Flow to the latest eventing library version `0.2` (version `0.2.2`).

#### Changelog

**New**

- `lib.eventing:0.2.2`
- Improved stability.
- Support for multiple streams and consumers.
- `EventingSubscriberClient` which is responsible for subscribing and listening for new messages from NATS Jetstream.

**Changed**

- If `eventing.enabled` property in `application.properties` is set to `false`, other eventing-related properties are not required in the file (and does not crash the service from launching itself if these properties are omitted completely from the file).
- Clean-up for `pom.xml` and various other classes.
- Various typos.

**Removed**

- Old `JetstreamListener` class and any related classes, objects and files (clean-up).

#### Testing / Reviewing

The service was tested e2e with the deployed NATS server in DEV OCP2. The `flow.service.listener`  was running locally and connected to remote NATS server. The `flow.service.workflow` was also running locally, but connected to all the required components remotely (i. e. `bmrg-core-services-users`, `bmrg-flow-services-controller`, `bmrg-dev-mdb001-mongodb` and `bmrg-dev-nats`). All tests were executed by launching a custom-defined workflow through Cloud Events:

- CE -> `service.listener` -> NATS Jetstream <-> `service.workflow` -> execute workflow

#### Note

This PR goes hand in hand with:

- boomerang-io/charts#46
- boomerang-io/flow.service.listener#16

PR #145, opened 2022-01-21
Feat 241 multischeduletriggers

corcoja:
This PR contains the upgraded eventing features that were merged into 3.4 (#126) but for some reason are not present in 3.5. I think it also includes (#132) but I'm not 100% sure since there are 2 unmerged branches for multi-schedule triggers:
image
image


tlawrie:
Hi @corcoja the real schedules branch is `feat-241-multischeduletriggers-3.5` which is already merged into `release-3.5`

Does that mean we don't need this PR? I am confused. If a branch is missing from `release-3.5` can we not merge directly?

corcoja:
> Does that mean we can instead merge `corcoja:feature/lib.eventing` into `release-3.5` or merge `release-3.4` into `release-3.5` to make sure it covers everything?

`lib.eventing` has been already merged into `release-3.4` (here #126), so I would go with merging `release-3.4` into `release-3.5`. I will open a new PR for this.

corcoja:
There were some unpruned branches in my forked repo and this confused me. `release-3.4` has been already merged into `release-3.5` so it's all good. I have just double checked the code for `lib.eventing` and everything looks to be in place. Closing this PR.

tlawrie:
Awesome. Thanks for checking @corcoja. I actually was worried about this happening.

PR #147, opened 2022-01-24
fix create workspace

amhudson:
Closes #

{{short description}}

#### Changelog

**New**

- N/A

**Changed**

- Update createWorkspaceRequest properties

**Removed**

- N/A

### DIFF
--- a/src/main/java/io/boomerang/service/runner/misc/ControllerClientImpl.java
+++ b/src/main/java/io/boomerang/service/runner/misc/ControllerClientImpl.java
@@ -123,7 +123,7 @@ public class ControllerClientImpl implements ControllerClient {
 
 
     if (enableStorage) {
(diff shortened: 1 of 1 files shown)


tlawrie (changes requested):
Hi @amhudson i think at this point it was correctly creating the **Activity Workspace** but the code in `createWorkspaceRequest` was incorrectly referencing workflow instead of activity

tlawrie:
Thanks @amhudson lets give that a shot.

Can we double check whereever the Workflow Workspace gets created. I think its in the Save on the configure screen. Make sure its pulling in the defaults for the Workflow Workspace settings?

tlawrie (approved):

tlawrie:
Change looks good. Discussed on Slack

PR #158, opened 2022-03-08
adding annotations for workflow activity status changes

amhudson:
Closes #

{{short description}}

#### Changelog

**New**

- {{new thing}}

**Changed**

- {{change thing}}

**Removed**

- {{removed thing}}

#### Testing / Reviewing

### DIFF
--- a/src/main/java/io/boomerang/service/refactor/TaskServiceImpl.java
+++ b/src/main/java/io/boomerang/service/refactor/TaskServiceImpl.java
@@ -536,11 +536,14 @@ public class TaskServiceImpl implements TaskService {
     boolean workflowCompleted = dagUtility.validateWorkflow(activity);
     
     if (activity.getStatusOverride() != null) {
+      //Workflow activity changed to the override status
       activity.setStatus(activity.getStatusOverride());
     } else {
       if (workflowCompleted) {
+        //Workflow activity changed to completed
         activity.setStatus(TaskStatus.completed);
       } else {
+        //Workflow activity changed to failure
         activity.setStatus(TaskStatus.failure);
       }
     }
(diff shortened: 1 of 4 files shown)


amhudson:
@corcoja The activity itself will not ever be in `waiting` status, however an individual task might be (https://ocp2.cloud.boomerangplatform.net/dev/flow/apps/flow/activity/621e32b434f5f33fc28e7e33/execution/62278cc4b3dc8a14149de77e for example) and the activity status will be `inProgress`. I think we should publish event for updates to the workflow activity tasks as well as the workflow activity. If so, I can add comments to annotate the task status updates.

corcoja:
> @corcoja The activity itself will not ever be in `waiting` status, however an individual task might be (https://ocp2.cloud.boomerangplatform.net/dev/flow/apps/flow/activity/621e32b434f5f33fc28e7e33/execution/62278cc4b3dc8a14149de77e for example) and the activity status will be `inProgress`

Hey @amhudson, that's good to know! I think this won't be an issue since a workflow that is waiting for an WFE is, theoretically, still in progress.

> I think we should publish event for updates to the workflow activity tasks as well as the workflow activity. If so, I can add comments to annotate the task status updates.

Sounds like a good idea 👍 For now, I want to focus on the status of the workflow first and make it work. Afterwards, we can analyze the amount of messages that are being generated as well as the business need for activity statuses and decide on this.

src/main/java/io/boomerang/service/refactor/TaskServiceImpl.java
```diff
@@ -536,11 +536,14 @@ private void finishWorkflow(ActivityEntity activity) {
     boolean workflowCompleted = dagUtility.validateWorkflow(activity);
     
     if (activity.getStatusOverride() != null) {
+      //Workflow activity changed to the override status
```
corcoja, line 539: What "override" status exactly means? Have never seen this in the UI.

amhudson: statusOverride is applicable for the `Set Workflow Result Status` task template

amhudson:
> Sounds like a good idea 👍 For now, I want to focus on the status of the workflow first and make it work. Afterwards, we can analyze the amount of messages that are being generated as well as the business need for activity statuses and decide on this.

I'll  keep the PR as it is.

corcoja (approved):
Looks good. Thanks @amhudson!

PR #172, opened 2022-04-01
[ISEEC-3635] Persist the info of the Cloud Event that triggered the w…

AndreiOla:
**New**
 Persist the initiator and context of the Cloud Event that triggered the workflow, so we know which app triggered the event and some details about it.


tlawrie:
Hi @AndreiOla thanks for the PR. Is there an associated issue? Also could you fill in the detail to the PR?

The design for storing the triggering event is that we stored them in labels on the workflow activity.

I'll have to take a deeper look to determine the difference between the current design and this PR.

corcoja:
Hey @tlawrie. This PR is part of the workflow activity status Cloud Events that will be produced by Flow when eventing is enabled. The purpose of the new `eventProperties` property is to store additional details about a certain workflow activity that was triggered by a Cloud Event. Those properties will be extracted from the Cloud Event payload then added to `ActivityEntity` collection.

@AndreiOla, @amhudson and myself decided to store this information inside the `ActivityEntity` collection for the following reasons:
1. Have this information stored just once for a single activity instead of having multiple entries for each task individually as Tekton `taskrun` labels.
2. We didn't want to add any eventing related information to the Tekton `taskrun` object, since eventing is an addition to the platform and is not required to have Flow functioning without it.

The details that we're planning to store as `eventProperties` are below (note that these are optional and are not required by any means):
1. `initiator_id` - different from `source`. This ID will be used to generate the NATS subject of the workflow activity status Cloud Events. By having a "custom" NATS subject, clients consuming these messages can filter it by the `initiator_id`.
2. `context` - some information about the Cloud Event context (different from Flow parameters). Someone who has triggered a workflow with a Cloud Event and has also added the `context` property inside the Cloud Event, will receive the same `context` property when consuming workflow activity status Cloud Events.

Please let us know if you think otherwise or have a better suggestion where to store this information 😊

Also, please note this is a PR for the new eventing branch, which is still WIP.

tlawrie:
Hi @corcoja thanks for the detail and the clarifications. Its always so well laid out! 

Let me respond tomorrow with more detail to clarify what I was thinking as an alternative.

tlawrie:
Hey, @corcoja please find the following to clarify my prior question/statement that explains the overall direction we are going with to get to 4.0

**Labels** 
To clarify, I wasn't meaning the TaskRuns or a Kubernetes object, instead, I was referring to the Flow labels which have slowly been becoming a part of the structure with the intent that by the 4.0 architecture they will be a first-class citizen.

The idea is that in the UI or via the APIs you will be able to query for Activity / Workflows / Users / Teams / etc with labels (the same as you can do in Kubernetes and a number of cloud-native Kubernetes based systems). A number of these Flow objects support labels:
- Workflows / Users / Teams you can create and view labels in the UI. 
- Activity you can query labels via the API.

One of the purposes for labels is exactly for things like this, additional metadata without needing to create another unique extension to the object, and therefore have the flow on maintenance i.e. update the implementation.

You can find examples of this in the `EventProcessorImpl.java:195-200`.

**Proposed Design Adjustment**
1. Leverage the 'labels' on an activity to store the CloudEvent based metadata.
2. Store as many of the CloudEvent elements as labels (or even all of the standard payload/header metadata)
3. Determine whether to use a `ce-xyz=value` label or `io.boomerang.event/xyz=value`. The second example matches the prefix used in the Event but also is a standard Kubernetes way of creating labels for a system.

**CloudEvents Payload**
An additional area of discussion is as you mentioned the eventProperties which are not part of the [CloudEvents specification](https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/spec.md). The spec is relied upon, and used mostly in the same way, by IBM / Google / Microsoft etc so we don't want to alter the course too much from the proven

I'm also not sure `initiator_id` or `context` were discussed in the [original GitHub Issue](https://github.com/boomerang-io/roadmap/issues/326) regarding the enhancement. Are we saying these are extensions to the CloudEvent payload or generated in the code for NATS specifically?

corcoja:
Hey @tlawrie,
Thanks a lot for the detailed explanation of labels. We will need to look more into labels and get acquainted with the concept in order to adapt our requirements (@AndreiOla). Of course, our goal is to leverage as much as possible from what the platform already offers, so this is useful information to us.

> I'm also not sure initiator_id or context were discussed in the https://github.com/boomerang-io/roadmap/issues/326 regarding the enhancement.

I have not described these two attributes there since this is something we identified we need while designing the architecture for "updated" eventing.

> Are we saying these are extensions to the CloudEvent payload or generated in the code for NATS specifically?

These are indeed an extension to the CloudEvent payload, and while we don't want to deviate from the spec, we couldn't identify any specific place/field where we could assign these attributes (this is excluding the `data` field which is used for workflow parameters). @AndreiOla is looking into CloudEvent v3 which allows the extension of the default CloudEvent with additional information.

AndreiOla:
Hi @tlawrie @corcoja I've researched a little bit deeper this issue and I think the best solution would be to create 2 new extension attributes, that follow the naming conventions for CE. Basically I compiled this information from 2 sources:
https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/extensions/distributed-tracing.md
https://w3c.github.io/trace-context/
We cannot use the Known Extension: Distributed Tracing as this would be tied to the corresponding info in Headers (for a HTTP bound CE message) and also the fields are too strict and not necessarily related to what we want.
But based on those fields I think we should define our own:

**initiatorid**
        Type: String
        Description: This describes the highest level of the event producer, because the `subject `and `source` is not sufficient to      identify it. Typically we have an app that is sending CEs with different `subjects` and with different `sources`. We want to quickly identify the app that sends all these CEs, without the need to filter all the messages received and figure it out by the subject and source. 
The **initiatorid** should be created as a 16 HEX string, more like a WWN, so we can avoid duplicate values, but this should not be enforced: https://en.wikipedia.org/wiki/World_Wide_Name
**initiatorcontext**
         Type: String
         Description: This may contain more information about the app that initiated the CE.

Example:
{
    "specversion" : "1.0",
    "type" : "com.github.pull_request.opened",
    "source" : "https://github.com/cloudevents/spec/pull",
    "subject" : "123",
    "id" : "A234-1234-1234",
    "time" : "2018-04-05T17:31:00Z",
    "initiatorid" : "4bf92f3577b34da6",
    "initiatorcontext" : "context description of the initiating app",
    "datacontenttype" : "text/xml",
    "data" : "<much wow=\"xml\"/>"
}

tlawrie:
Hi @corcoja @AndreiOla that was a super-fast response! And thank you for elaborating on the reasoning and an example payload. Really helped me understand where the idea is coming from.

**Labels**
The addition of labels was started in the 3.x versions of Flow and had been a major design choice, and prior to me moving on. I imagine its been lost over successive handovers. Ill create a specification and add it to [Flows Architecture](https://github.com/boomerang-io/architecture/tree/main/flow).

**Use of Extensions**
We can create extensions, and we have for internal use between the listener and the workflow service. You can see the implementation in the [Listener Service's EventProcesserImpl](https://github.com/boomerang-io/flow.service.listener/blob/main/src/main/java/io/boomerang/service/EventProcessorImpl.java) lines 115-122.'

I agree that, if these two additional fields (initiatorid and initiatorcontext)  are needed, that extensions are the right way to go.

**Eventing Payload Design**
To clarify on the need for these additional fields,
1. Its mentioned that the source is not sufficient. We followed the design that IBM/Google/Microsoft use which is contextually specific to the object type or path for an API driven system. The example given for events from bucket storage would be `/catalog/services/cloud-object-storage/mybucket` or `/catalog/services/cloud-object-storage/mybucket`. It should identify the source of the event. So in the scenario mentioned, where would the event be coming from that a context path like this wouldn't identify the end system?

2. If these extension fields are made 'optional' does the system still work? Will anything break if they are not provided?

**Updated Eventing Design**
1. Are we still planning to follow the design of a configured 'sink' to produce the events to? (goes to a broader design question but from my understanding is what these additional fields will be passed to). And if so, can it still be configured via settings and 'may' be a different NATs (or eventing system like Azure Event Grid, Google Eventarc) than the one used for incoming/triggering events.

2. Can we get the [original enhancement issue](https://github.com/boomerang-io/roadmap/issues/326) updated with any updates/alterations being proposed to the design.

AndreiOla:
Hi @tlawrie Here is my understanding of what is required here:
"1. Its mentioned that the source is not sufficient. We followed the design that IBM/Google/Microsoft use which is contextually specific to the object type or path for an API driven system. The example given for events from bucket storage would be /catalog/services/cloud-object-storage/mybucket or /catalog/services/cloud-object-storage/mybucket. It should identify the source of the event. So in the scenario mentioned, where would the event be coming from that a context path like this wouldn't identify the end system?"

There may be multiple event sources that are sent from the same app. So, there would be multiple `sources` . We want to identify the app. (E.g: We have App1, that is composed of multiple microservices: micro1, micro 2, etc., , so the `initiatorid` would be a HEX value that uniquely identifies App1). 

    "2. If these extension fields are made 'optional' does the system still work? Will anything break if they are not provided?"

These values would be retrieved from the CE that is received and then added as a suffix in the NATS subject of a CE. This way, when the CE goes through the NATS Stream, a Consumer can  filter only the events that are related to a certain App.
E.g: bla.bla.status.completed.App1 and the filter would be: bla.bla.status.*.App1 If no initiatorid was specified, this means that either the Activity was generated by Flow itself, or the CE sender did not want to persist it. And the events would go in the NATS Stream with the Subject missing the final suffix. So, it is the Consumer's job to figure out what to do with those CEs.

tlawrie:
Ok. So it sounds like
1. It wouldn't break anything.
2. We would use CE extensions.
3. We would use labels as the storage mechanism for an activity

However;
- I am not sure I understand the source issue, ultimately. Should the source not be able to identify it, if the path has been narrowed enough? i.e. if its App1, the source could be `/catalog/services/App1/myVersionOfApp1`? Could we drill down into some more detail around what a real source would look like in the current design. Then we can compare that to the various main implementations and see what it 'could' be to solve the issue.

- If we do implement an initiatorId, i'd suggest that we would always want it to be both fitlerable and also go to non filtered as per this section of the diagram in @corcoja design diagram
image

- If I look at other 'sink' type implementations, typically you can extend a system to output the events to a specific (or multiple) sink(s). In our implementation, we are starting with the system (again I reiterate here that the system's NATs may be different than the configured inbound NATs) and then eventually Team based configured sinks. I just want to confirm and make sure that its 
  - still configurable
  - Works regardless of the extensions.

AndreiOla:
1. These extensions are Optional. Since they are optional, it is the publisher's and consumer's job to decide if these values have any meaning for them or not.
2. We need to have a static value that identifies the _App1_ from our example and not something that is inferred/substracted from other fields, like `source` or 'subject' that may be changed by the event publisher.

tlawrie:
Alright. Well in that case. Are we able to confirm

1. It wouldn't break anything.
2. We would use CE extensions for `initiatorid` and `initiatorContext`
3. We would use labels as the storage mechanism for an activity.
4. We would use a prefix for the event specific labels. Such as `io.boomerang.event/xyz=value`
5. The output sink is configurable (to start with at the system level, eventually per team). For a number of reasons, including that you may be using a cloud provider's SaaS service such as eventarc or eventgrid and therefore need to output to a different location. You may have incoming events disabled but still want Flow to output events. It also is standard practice with the compared projects previously mentioned.

tlawrie:
Added the [labels specification](https://github.com/boomerang-io/architecture/blob/main/flow/Labels.md) and added it to [Flows Architecture](https://github.com/boomerang-io/architecture/tree/main/flow).

AndreiOla:
Hi @tlawrie These 'labels' looks like a solution for our problem.
But I need to understand where are the labels stored in the context of a CE message:
{
"specversion" : "1.0",
"type" : "com.github.pull_request.opened",
"source" : "https://github.com/cloudevents/spec/pull", ------> **io.boomerang.event/source**
"subject" : "123",
"id" : "A234-1234-1234", -----> **io.boomerang.event/id**
"time" : "2018-04-05T17:31:00Z",
"initiatorid" : "4bf92f3577b34da6", ------> **io.boomerang.event/initiatorid**
"initiatorcontext" : "context description of the initiating app", ------> **io.boomerang.event/initiatorcontext**
"datacontenttype" : "text/xml",
"data" : "<much wow="xml"/>"
}
It does not seem to respect the CE Spec if we do that, since there are some  Interface methods there.

I does not really matter if the attribute has a prefix before it, as long as its spec does not change. (i.e: If I get a 'prefix/initiatorid', I know what was the initiating App, that sent the CE). It is way more precise than to look for some substring in the 'source' attribute, for example.

tlawrie:
Hi @AndreiOla the labels don't affect the CE Spec. You would use the extensions. It's how they are stored on the activity

**CE Spec Exampe**
```JSON
{
"specversion" : "1.0",
"type" : "com.github.pull_request.opened",
"source" : "https://github.com/cloudevents/spec/pull",
"subject" : "123",
"id" : "A234-1234-1234",
"time" : "2018-04-05T17:31:00Z",
"initiatorid" : "4bf92f3577b34da6",
"initiatorcontext" : "context description of the initiating app",
"datacontenttype" : "text/xml",
"data" : "<much wow="xml"/>"
}
```

**Flow Storage on Activity - FlowExecutionRequest.java & ActivityEntity.java**
```JSON
{
    "_id" : ObjectId("61d628ccf3bc25396a5fb2fa"),
    "scope" : "user",
    "workflowRevisionVersion" : NumberLong(0),
    "creationDate" : ISODate("2022-01-05T23:25:00.433Z"),
    "status" : "cancelled",
    "workflowId" : "61d620a33aa9034ded32caee",
    ...
    "labels" : [{
       "key": "io.boomerang.event/initiatorid"
       "value": "4bf92f3577b34da6"
    },{
       "key": "io.boomerang.event/initiatorcontext"
       "value": "context description of the initiating app"
    }]
}
```

AndreiOla:
Right now, I persist the   initiatorid and initiatorConntext as private 'Map<String, String> eventProperties'. (Maybe find a better name like initiatorEventProperties? ), in ActivityEntity.
If there is a reason to persist these values in  private 'List<KeyValuePair>' labels, there is not big change to do it so.

tlawrie:
Hi @AndreiOla yes, I believe that's what we have been discussing above, is that the Labels is the perfect place to store this additional metadata (and any future extensions to the object) in that its purpose-built for that without needing to adjust / add on / extend the entities model (and true for all the other objects that we have standardized to labels such as Teams, Users, Workflows, etc)

tlawrie:
Hi @AndreiOla how are you going with this?

AndreiOla:
Hi @tlawrie I've been on vacation. That makes sense, to store those Extensions as Labels. I will make the changes, accordingly.

AndreiOla:
@tlawrie  I have question: Where are we getting these prefixes from?
"key": "**io.boomerang.event**/initiatorcontext"
For now I will hardcode this prefix. But maybe we need a common class, from where to get these shared prefixes for labels?

tlawrie:
Hi @AndreiOla its the prefix we use for the events throughout the system and the labels in the helm chart and anywhere we need a reverse DNS based context and matches the industry with how the prefixes are implemented in cloud-native objects.

You can have it come from an application.properties if you want. I think the listener service currently has it for the events it passes on.

AndreiOla:
The code resulted from this PR was integrated into branch [feature/eventing_wf_status](https://github.com/boomerang-io/flow.service.workflow/tree/feature/eventing_wf_status).

PR #174, opened 2022-04-13
Fix/update spring

AndreiOla:
**Changed**
Updated Spring to 5.2.20.RELEASE, in order to fix critical vulnerability, for tag 2.8.3.7 .

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -32,7 +32,72 @@
 		<jsonwebtoken.version>0.10.7</jsonwebtoken.version>
 		<jackson.databind.version>2.9.9.3</jackson.databind.version>
 		<log4j2.version>2.17.0</log4j2.version>
+		<spring.version>5.2.20.RELEASE</spring.version>
 	</properties>
+	<dependencyManagement>
+	<dependencies>
+		<dependency>
+			<groupId>org.springframework</groupId>
+			<artifactId>spring-web</artifactId>
+			<version>${spring.version}</version>
+		</dependency>
+		<dependency>
+			<groupId>org.springframework</groupId>
+			<artifactId>spring-tx</artifactId>
+			<version>${spring.version}</version>
+		</dependency>
+		<dependency>
+			<groupId>org.springframework</groupId>
+			<artifactId>spring-core</artifactId>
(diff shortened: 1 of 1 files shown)


tlawrie:
Hi @AndreiOla this seems to have a number of other changes not related to the dependency change. Could you clarify?

AndreiOla:
Hi @tlawrie I made changes starting from tag 2.8.3.7. And I want to build only those changes, on top of tag 2.8.3.7. I was just trying to figure out how can I do that, since git does not allow me to open a PR against a certain tag.

tlawrie:
So typically you would create a branch from the tag i.e. 

```
git checkout <tag>
git checkout -b fix-<branch_name>
```

And then when you tag it, it would be the original <tag> and then either a fourth hot fix number like `.1` or `-fix.1` or something like that

AndreiOla:
Thanks @tlawrie !
I made separate branches from the tags in my fork, made changes, and then used @amhudson branches in (https://github.com/boomerang-io/flow.service.workflow), for the tag, since I do not have sufficient rights to push into (https://github.com/boomerang-io/flow.service.workflow) directly.
That is correct, that we don't need to merge into main. Just build hot fixes for those tags.

amhudson:
@tlawrie yes, that's correct

pom.xml
```diff
@@ -31,8 +31,74 @@
 		<jacoco.version>0.8.2</jacoco.version>
 		<jsonwebtoken.version>0.10.7</jsonwebtoken.version>
 		<jackson.databind.version>2.9.9.3</jackson.databind.version>
+		<log4j2.version>2.17.0</log4j2.version>
+		<spring.version>5.2.20.RELEASE</spring.version>
 		<log4j2.version>2.17.2</log4j2.version>
```
amhudson, line 36: there are 2 `log4j2.version`, remove `2.17.0`.

amhudson (changes requested):

pom.xml
AndreiOla: Fixed.

amhudson (approved):

PR #177, opened 2022-05-02
Iseec 3837

amhudson:
Closes #

{{short description}}

#### Changelog

**New**

- Adding endpoint to retrieve a workflow token and WFE topic value to manually trigger a WFE task

**Changed**

- N/A

**Removed**

- N/A

#### Testing / Reviewing

Tested and verified locally

### DIFF
--- a/src/main/java/io/boomerang/controller/InternalController.java
+++ b/src/main/java/io/boomerang/controller/InternalController.java
@@ -116,7 +116,7 @@ public class InternalController {
     return configurationService.updateSettings(settings);
   }
 
-  @GetMapping(value = "/activity/triggerWFE")
+  @GetMapping(value = "/activity/revision-properties")
   public ResponseEntity<WFETriggerResponse> getRevisionProperties(@RequestParam String workflowId,
       @RequestParam long workflowVersion, @RequestParam String taskId,
       @RequestParam String propertyKey) {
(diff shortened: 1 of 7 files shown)


tlawrie:
Hey @amhudson could you provide a Git Issue or some context in the PR for this one?

Also a few clarification questions
1. Is this adding a new internal endpoint? Is there a way we can make it an API call protected by an API token (maybe Global and/or Team scope)? The reason is that we are moving to deprecate all internal endpoints so there are only the authenticated API endpoints or eventing endpoints.

2. Its a GET endpoint but called triggerWFE. Is there a better name if its just a GET?

3. Is there a security concern with responding back with a Token? Also it seems to just be getting the first Token... which may not be the token that has been set up to be used for the WFE. The reason we have multiple tokens on a workflow is so that they can be named appropriately and used in a narrow scope i.e. creating a token specifically for WFE or specificially for an Integration like Slack so that if you have to revoke one it doesn't impact the rest.

amhudson:
We are trying to give privileged users the ability to trigger a WFE manually. This endpoint is retrieving the properties to allow us to do so. 

1. We currently still use the internal endpoints especially during automated request processing. We are very much behind the latest Flow release and we have not made any provisions with integrating API tokens. 
2. Yep, I will rename
3. At the moment, we just need any workflow token in order to manually trigger the WFE. 

@gchickma thoughts?

tlawrie:
Thanks for the context @amhudson.  I can see if you have lots of WFE it would be a pain to handle.

I recommend over time this migrate from an internal endpoint to a true API, or we even build functionality to support the resolution of WFE in Flow (almost like an Action)

tlawrie (approved):

PR #180, opened 2022-05-04
revert dependency updates

amhudson:
@AndreiOla https://github.com/spring-projects/spring-shell/issues/324

Updating to spring boot 2.6.x is causing similar issues. We need to revisit this upgrade and thoroughly test any updates for issues.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -12,7 +12,7 @@
 	<parent>
 		<groupId>org.springframework.boot</groupId>
 		<artifactId>spring-boot-starter-parent</artifactId>
-		<version>2.6.6</version>
+		<version>2.5.3</version>
 		<relativePath />
 	</parent>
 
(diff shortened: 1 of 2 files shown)


tlawrie (approved):

AndreiOla:
@tlawrie @amhudson In Spring 2.6.x : Circular References Prohibited by Default! (The issue was there before, but it was hidden!)
The workaround is to set the application property that I set.
The fix implies a refactoring: https://www.baeldung.com/circular-dependencies-in-spring
So, I would leave the workaround in place and plan for a refactoring, so at least we have the vulnerabilities fixed.

tlawrie:
Hi @AndreiOla im not following on the workaround.

Right now, the main branch and any tag with this code in it, doesn't work. Doesn't seem there is a workaround in place.

Can we not upgrade to the latest 2.5.x and leave it at that for now until the service is refactored?

AndreiOla:
@tlawrie I may have missed adding `spring.main.allow-circular-references=true` property also in application.properties. That should workaround.

PR #182, opened 2022-05-09
Added code to filter sensitive data from properties of workflows and …

AndreiOla:
…teams

Closes #
https://github.com/boomerang-io/roadmap/issues/336

### DIFF
--- a/src/main/java/io/boomerang/service/crud/TeamServiceImpl.java
+++ b/src/main/java/io/boomerang/service/crud/TeamServiceImpl.java
@@ -279,21 +279,32 @@ public class TeamServiceImpl implements TeamService {
 
     if (flowTeamEntity.getSettings() != null
         && flowTeamEntity.getSettings().getProperties() != null) {
-    	filterPasswordValue(flowTeamEntity.getSettings() == null ? null:flowTeamEntity.getSettings().getProperties());
+    	filterPasswordValue(flowTeamEntity.getSettings() == null ? null:flowTeamEntity.getSettings().getProperties(), true);
       return flowTeamEntity.getSettings().getProperties();
     } else {
       return Collections.emptyList();
     }
   }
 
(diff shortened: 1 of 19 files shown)


src/main/java/io/boomerang/service/crud/WorkflowServiceImpl.java
```diff
@@ -336,14 +337,25 @@ public WorkflowSummary updateWorkflowProperties(String workflowId,
...
       return new WorkflowSummary(entity);
     }
   }
 
+  private List<WorkflowProperty> filterPasswordValue (List<WorkflowProperty> properties) {
+	  final String passTypeStr = "password";
+	  if (properties == null) return null;
+	  //If the property is a password, do not return its value, for security reasons.
+		Optional<WorkflowProperty> passProp = properties.stream().filter(f -> passTypeStr.equals(f.getType()) && f.getDefaultValue() != null).findAny();
+		if (passProp.isPresent()) passProp.get().setDefaultValue(null);
```
tlawrie, line 355: Should this also do the value? or just the default value?

src/main/java/io/boomerang/service/crud/TeamServiceImpl.java
```diff
@@ -273,19 +279,45 @@ public List<FlowTeamConfiguration> getAllTeamProperties(String teamId) {
...
+			if (isDefaultValue) {
+				passProp.get().setDefaultValue(null);
+			} else {
+				passProp.get().setValue(null);
+			}
+		}
+		return properties;
+	}
+  
+
```
tlawrie, line 309: Can we make this a common util method in `io.boomerang.util` package that can be used by any impl over time? and make it so both team and workflow can use the same shared methods (even if there are two or three) it centralizes the filtering code.

src/main/java/io/boomerang/service/crud/WorkflowServiceImpl.java
AndreiOla: As far as I know Workflows use defaultvalue and Teams use value.

PR #200, opened 2022-06-10
adding support for dot and bracket parameter notation

amhudson:
Closes #

https://github.com/boomerang-io/roadmap/issues/345

#### Changelog

**New**

- Adding support for dot and bracket parameter notation

**Changed**

- N/A

**Removed**

- N/A

#### Testing / Reviewing

Tested and verified locally


tlawrie (commented):
Hey @amhudson i think lets do a session on this in the next Boomerang community call. I think we could also be changing this in the Parameter Manager

I've rewritten part of this in https://github.com/boomerang-io/flow.service.engine/blob/main/src/main/java/io/boomerang/service/ParameterManagerImpl.java

It still uses the paramLayers and builds them up. But i think overall we could solve this similar to how you have it.

PR #207, opened 2022-06-20
Initialize 'templateIds' when scope is template

### LINKED ISSUE #206: Fail to import Template workflows, Workflow not imported - template(s) not found
**Describe the bug**
Trying to import a workflow in the Template section fails with HTTP 400 and the message: 
```
Workflow not imported - template(s) not found
```

**To Reproduce**
1. Create any workflow with the existing standard tasks and export it,
2. Go to Administer -> System Workflows -> Templates
3. Create a new template by importing the workflow created at step 1. 

**Expected behavior**
The template should be imported.

**Environment (please complete the following information):**

- OS: Any OS
- Browser: Any browser
- Version: 3.9.0

morarucostel:
Closes #206 

The `templateIds` array is not initialized with the global and system taskids when the scope sent by the UI is `template`. 
Adding the initialization of `templateIds` array just like in the case of system imported workflows.

#### Changelog

**New**

- NA

**Changed**

- initialization of `templateIds` array just like in the case of system imported workflows

**Removed**

- NA

#### Testing / Reviewing

On local and dev environments.

### DIFF
--- a/src/main/java/io/boomerang/service/crud/WorkflowServiceImpl.java
+++ b/src/main/java/io/boomerang/service/crud/WorkflowServiceImpl.java
@@ -639,14 +639,14 @@ public class WorkflowServiceImpl implements WorkflowService {
 
   /*
    * Checks if the Workflow can be executed based on an active workflow and enabled triggers.
-   *
+   * 
    * If trigger is Manual or Schedule then a deeper check is used to check if those triggers are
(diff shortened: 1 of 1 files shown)


tlawrie:
Seems like a simple enough but good change.

tlawrie (approved):

PR #209, opened 2022-06-23
Develop new utility class for computing the Activity Status

AndreiOla:
**New**

I propose a solution on how to compute the resulting Activity status, based on each Tasks' Status that are executed in the corresponding workflow.

### DIFF
--- a/src/main/java/io/boomerang/util/ActivityStatusResolver.java
+++ b/src/main/java/io/boomerang/util/ActivityStatusResolver.java
@@ -40,15 +40,20 @@ public class ActivityStatusResolver {
     TaskStatus resultingTaskStatus = null;
     switch (decimalStatus) {
       case 3: // All Waiting Bit Array: 00000010
+      case 18: // Waiting, NotStarted
       case 66:// Waiting, Failure Bit Array: 01000010
+      case 82: // Waiting, NotStarted, Failure
       case 130:// Waiting, Complete
+      case 146: // Waiting, NotStarted, Complete
       case 194:// Waiting, Complete, Failure
+      case 210: // Waiting, Not Started, Complete, Failure
         resultingTaskStatus = TaskStatus.waiting;
         break;
       case 128: // All Complete
         resultingTaskStatus = TaskStatus.completed;
         break;
-      case 192: // All Complete OR Failure. At least one Complete
+      case 192: // All Complete OR Failure. At least one Complete and one Failure
+        // TODO Is this Complete or Failure?
         resultingTaskStatus = TaskStatus.completed;
         break;
       case 64:
--- a/src/main/java/io/boomerang/util/ActivityStatusResolver.java
+++ b/src/main/java/io/boomerang/util/ActivityStatusResolver.java
@@ -1,0 +1,84 @@
+package io.boomerang.util;
+
+import java.util.List;
+import org.springframework.beans.factory.annotation.Autowired;
+import io.boomerang.model.Task;
+import io.boomerang.mongo.model.TaskStatus;
+import io.boomerang.mongo.service.ActivityTaskService;
+
+/**
+ * @author O17006826 This class is used to generate a decimal number that represents the combined
+ *         Status to be returned for a certain list of Tasks' Statuses
+ */
+public class ActivityStatusResolver {
+  @Autowired
+  private ActivityTaskService taskActivityService;
+
+
+  public TaskStatus computeStatus(List<Task> taskList) {
+
+    // Each Status will have a bit that corresponds to its index position
+    // completed("completed"), 0
+    // failure("failure"), 1
+    // inProgress("inProgress"), 2
+    // notstarted("notstarted"), 3
+    // invalid("invalid"), 4
+    // skipped("skipped"), 5
+    // waiting("waiting"), 6
+    // cancelled("cancelled"); 7
+    Byte[] statusBitArray = new Byte[TaskStatus.values().length];
+    for (Task task : taskList) {
+      TaskStatus status = (taskActivityService.findById(task.getTaskId())).getFlowTaskStatus();
+      // If we find at least one TaskExec with status inProgress for ex, then set its corresponding
+      // bit to 1
+      statusBitArray[status.ordinal()] = 1;
+    }
+    return computeResultingStatus(convertbitArraytoDecimal(statusBitArray));
+  }
+
+  private TaskStatus computeResultingStatus(int decimalStatus) {
+    TaskStatus resultingTaskStatus = null;
+    switch (decimalStatus) {
+      case 3: // All Waiting Bit Array: 00000010
+      case 66:// Waiting, Failure Bit Array: 01000010
+      case 130:// Waiting, Complete
+      case 194:// Waiting, Complete, Failure
+        resultingTaskStatus = TaskStatus.waiting;
+        break;
+      case 128: // All Complete
+        resultingTaskStatus = TaskStatus.completed;
+        break;
+      case 192: // All Complete OR Failure. At least one Complete
+        resultingTaskStatus = TaskStatus.completed;
+        break;
+      case 64:
+        resultingTaskStatus = TaskStatus.failure;
+        break;
+      case 16: // All Not Started
+        resultingTaskStatus = TaskStatus.notstarted;
+        break;
+      case 1: // All Cancelled
+        resultingTaskStatus = TaskStatus.cancelled;
+        break;
+      // TODO Need to determine what other Statuses result from the Task Status combinations
+      default:
+        resultingTaskStatus = TaskStatus.inProgress;
+        break;
+    }
+    return resultingTaskStatus;
+  }
+
+  private int convertbitArraytoDecimal(Byte[] statusBitArray) {
+    int decimal = -1;
+    StringBuilder binaryString = new StringBuilder();
+    for (Byte bit : statusBitArray) {
+      binaryString.append(bit);
+    }
+    try {
+      decimal = Integer.parseInt(binaryString.toString(), 2);
+    } catch (NumberFormatException nex) {
+      // ignore
+    }
+    return decimal;
+  }
+}


AndreiOla:
A workflow can have multiple Tasks that are in different Statuses, and that run in parallel on different 'branches'. So, in order to find out what Status to set for the whole workflow, we need to get the status for every dependent Task and check if it is Waiting, Complete, Failure, etc and, depending on the Statuses of all these Tasks, to compute if the overall Status of the workflow is Waiting, Complete, etc.
For example, if we want to send a WaitForEvent, the dependent Tasks need to have either All a Status of Waiting or at least one Status Waiting and all the others to be in Complete or Failure and not InProgress or NotStarted, etc.

corcoja:
> For example, if we want to send a WaitForEvent, the dependent Tasks need to have either All a Status of Waiting or at least one Status Waiting and all the others to be in Complete or Failure and not InProgress or NotStarted, etc.

@AndreiOla, you still might have `NotStarted` tasks when the workflow is in a `Waiting` state. `NotStarted` just means that the task hasn't started **yet** and it can't start since there are other tasks in front of it that are in a `Waiting` state.
Everything else in your statement is correct.

tlawrie:
Will need @marcusdroy to review as codeowner.

tlawrie:
Hi @AndreiOla few questions / comments

1. Can you Edit the summary of the PR and add context around how its done now vs what is proposed? because as far as I know, workflows already have a status.
2. Can we simplify this and not use the bit assign. Just use a case on the actual status. Just do an array contains x status. This is applying some arbitary assignment and then doing conversion which imo is just confusing and unnecessary.

amhudson:
> Hi @AndreiOla few questions / comments
> 
> 1. Can you Edit the summary of the PR and add context around how its done now vs what is proposed? because as far as I know, workflows already have a status.
> 2. Can we simplify this and not use the bit assign?. Just use a case on the actual status. Just do an array contains x status. A utility class is still good, we have been trying to migrate anything shared and useful to a utility class, so thats great, but imo this is applying some arbitary assignment and then doing conversion which is confusing and unnecessary.

Additionally, can we keep the scope of these PRs to what is absolutely necessary to support eventing? There is no need to re-invent the wheel to do what we already do with existing code.

tlawrie:
Hi @corcoja @AndreiOla i think I get what you are trying to do. 

**Background**
Previously we left Waiting out of the status list. Part of it was 
- not wanting to do the overhead of being able to differentiate Waiting vs In Progress, but also 
- that Tekton Pipelines don't have a Waiting status for when we shift the engine completely to Tekton.

**Concern 1**
Concern around complexity/readability of this utility class for future maintainability with mappings between TaskStatus and bits.

**Concern 2**
What happens if a Task Fails, but then on the path we say continue on fail (i.e. fail is acceptable or good)? then the Task Statuses would include `TaskStatus.failure` yet it wouldn't be a failure of the Workflow... so it would just have to be deemed _In Progress_ or _Waiting_

As a TaskStatus alone is not equal to Workflow Status. You have to account for the DAG and link state + there is also a Task that sets Workflow Status to override the result.

You almost don't care about any of the other statuses because if the Workflow is marked as canceled or failed or completed, then thats what you just use as the ending status.

> **Thus this Utility is really only for a running workflows status.** and you accept the Workflow Status for any of the other statuses

**Alternative 1**
Based on concern 2 above, could we not just do a streamed array for distinct status and then, if it contains particular status of in progress or none in_progress && waiting then thats what the status is... any rely on the overall Workflow Status once finished to handle the Failure vs Success vs Cancelled vs Invalid.

This way we don't have to do any arbitrary bit number mapping and instead can focus on readable / consumable status types. Which also makes it easier to extend if we add another status.

_Pseudocode_
```java
    //Can be tested with a task status list such as
   // List<TaskStatus> runningStatuses = Arrays.asList(TaskStatus.notstarted, TaskStatus.notstarted, TaskStatus.notstarted, TaskStatus.completed, TaskStatus.waiting);

    public TaskStatus computeStatus(List<Task> taskList) {
      //Build up current task statuses
      List<TaskStatus> statuses = new LinkedList<>();
      for (Task task : taskList) {
        statuses.add(taskActivityService.findById(task.getTaskId())).getFlowTaskStatus())
      }

      if (statuses.contains(TaskStatus.inProgress)) {
        LOGGER.debug("Workflow is in Progress"); 
        return TaskStatus.inProgress;
      } else if (!statuses.contains(TaskStatus.inProgress) && statuses.contains(TaskStatus.waiting)) {
        LOGGER.debug("Workflow is Waiting"); 
        return TaskStatus.waiting;
     }
```

**Alternative 2**
A distinct array of TaskStatuses and use retainAll as a check compared against known states. This is a bit more complex and I feel we are going to run into issues that are stated in Concern 2 above.

_Pseudocode_
```java
//Waiting, Not Started, Complete, Failure
  private static final List<TaskStatus> waitingTaskStatuses = Arrays.asList(TaskStatus.waiting, TaskStatus.notstarted, TaskStatus.completed, TaskStatus.failure, TaskStatus.skipped);
  
  //In Progress, Waiting, Not Started, Complete, Failure, Skipped
  private static final List<TaskStatus> inProgressStatuses = Arrays.asList(TaskStatus.waiting, TaskStatus.notstarted, TaskStatus.completed, TaskStatus.failure, TaskStatus.inProgress, TaskStatus.skipped);

    public TaskStatus computeStatus(List<Task> taskList) {
      //Build up current task statuses
      List<TaskStatus> statuses = new LinkedList<>();
      for (Task task : taskList) {
        statuses.add(taskActivityService.findById(task.getTaskId())).getFlowTaskStatus())
      }
    
    List<TaskStatus> distinctStatuses = runningStatuses.stream().distinct().collect(Collectors.toList());
    LOGGER.debug("Is it waiting: " + distinctStatuses.retainAll(waitingTaskStatuses));   
    LOGGER.debug("Is it waiting: " + distinctStatuses.retainAll(inProgressStatuses));   
  }
```

tlawrie:
Further and in summary to the above comment

My recommendation is to 
1. Leave the ultimate ending status as whatever is set by the Workflow Execution. A final Workflow Status event can be sent at the end of the Workflow Execution

2. Only use a helper utility whilst the workflow is not yet completed / cancelled / failed / invalid. And as such you are really only checking to see if it is In Progress or Waiting.

corcoja:
@amhudson, the application clearly doesn't behave like this though.

Screen Shot 2022-06-30 at 5 49 40 PM

My understanding is that a workflow **can** have the Waiting status, which means that the execution is currently blocked by a Waiting task, and that is what @AndreiOla is implementing – the process to identify if a workflow is indeed Waiting or not.

amhudson:
> @amhudson, the application clearly doesn't behave like this though.
> 
> Screen Shot 2022-06-30 at 5 49 40 PM
> 
> My understanding is that a workflow **can** have the Waiting status, which means that the execution is currently blocked by a Waiting task, and that is what @AndreiOla is implementing – the process to identify if a workflow is indeed Waiting or not.

There should not be a Waiting tab there for the reasons Tyson previously mentioned.

> Previously we left Waiting out of the status list. Part of it was
>- the overhead of being able to differentiate Waiting vs In Progress, but also
>- that [Tekton Pipelines don't have a Waiting status](https://tekton.dev/docs/pipelines/pipelineruns/#monitoring-execution-status) for when we shift the engine completely to Tekton.

corcoja:
WFE task is not a task backed by a Tekton task, so I don't see why Tekton is a concern here?

amhudson:
> WFE task is not a task backed by a Tekton task, so I don't see why Tekton is a concern here?

image
The statuses here are **_workflow_** statuses, not **_task_** statuses. There should not be a **_Waiting_** status for a **_workflow_** . What this PR is trying to do is determine when to set a Waiting **_workflow_** status.

corcoja:
> There should not be a **_Waiting_** status for a **_workflow_**

Is there an open issue for this or PR? I can't find anything related to this either in the docs or in the roadmap project. If this was a verbal agreement between the open source community, it must be properly documented or at least open an issue to remove support for Waiting status for workflows.

Now, on the topic for Waiting statuses, here's my understanding:

1. A workflow can't be in Waiting. Why?
2. A WFE task though, which is not a Tekton task, can be in Waiting.

I don't see any technical impediments why we wouldn't be able to implement the Waiting status for a Workflow, since, from a business point of view, a workflow can be in Waiting status.

tlawrie:
So to summarise where we are at, so that progress can be made.

1. There is an objection to the need for the bit comparison method and a request to simplify the implementation.

2. There is discussion on the use of 'Waiting' status overall vs Task
  - I _think_ there is agreement that It does _logically_ make sense to have a **Waiting status on a Workflow**, _however_ this then highlights an issue with (3) below.

3. Question on the TektonCD alignment and changes
  - _Historical Reference:_ The alignment to TektonCD started with Tasks.The evolution of migrating to full Workflow (Pipeline) migration are part of the next architecture (which is currently in 4.0). At the time the first steps were taken with 3.0, the features in Tekton were not yet in full availability and were still in 'experimental' status, namely CustomTasks (for reference, that's the terminology we used in 1.x and 2.x for BYO tasks), onError, and a few other items. The overall eventual aim is to remove the need for any DAG processing to occur in the Workflow service. Instead, the ultimate aim is to have Tekton process it and Flow gain the benefit of a much larger ecosystem and bring in new features like Matrix Loops, Editing Workflows in YAML, etc.
  - _What does this mean:_ Eventually, we won't understand the 'Waiting' status for a Workflow. It will just be 'In Progress'. So its not something we can count on thats why work was stopped on trying to make Waiting work in all scenarios and for future efforts.

I believe the above captures where we are at with everything, now we need to discuss next steps on how to proceed with the above known.

AndreiOla:
@tlawrie Now it makes more sense, if there was an intention to migrating the code. My Task is to find out when a workflow is in Waiting State and send a WaitForEvent event. Not sure if we should still set he actual Workflow Status to Waiting, but anyway this can be done without any utility class and just use a method. I will leave this branch as is, so we keep the conversation and make a new branch with just the changes needed for the Task I have.

AndreiOla:
@tlawrie Now it makes more sense, if there was an intention to migrating the code. My Task is to find out when a workflow is in Waiting State and send a WaitForEvent event. Not sure if we should still set he actual Workflow Status to Waiting, but anyway this can be done without any utility class and just use a method. I will leave this branch as is, so we keep the conversation and make a new branch with just the changes needed for the Task I have.

corcoja:
@tlawrie, I have added a sample CE payload for task updates. See here: https://github.com/boomerang-io/roadmap/issues/326#issuecomment-1187375669

PR #210, opened 2022-06-27
Workflow service – Produce CloudEvents with workflow status

corcoja:
Part of feature request boomerang-io/roadmap#326

#### Changelog

**New**

- {{new thing}}

**Changed**

- {{change thing}}

**Removed**

- {{removed thing}}

#### Testing / Reviewing


src/main/java/io/boomerang/service/FlowExecutionServiceImpl.java
```diff
@@ -171,6 +168,8 @@ public void prepareExecution(List<Task> tasks, String activityId) {
       activityEntity.setStatus(TaskStatus.invalid);
       activityEntity.setStatusMessage("Failed to run workflow: Incomplete workflow");
       activityService.saveWorkflowActivity(activityEntity);
+      eventingService.publishActivityStatusEvent(activityEntity);
+
```
tlawrie, line 172: Can we also make this Async?

src/main/java/io/boomerang/controller/InternalController.java
```diff
@@ -70,14 +70,23 @@ public List<WorkflowShortSummary> getAllWorkflows() {
...
   @PutMapping(value = "/workflow/event", consumes = "application/cloudevents+json; charset=utf-8")
-  public ResponseEntity<CloudEventImpl<EventResponse>> acceptEvent(
-      @RequestHeader Map<String, Object> headers, @RequestBody JsonNode payload) {
-    return ResponseEntity.ok(eventProcessor.processHTTPEvent(headers, payload));
+  public ResponseEntity<EventResponse> acceptEvent(@RequestHeader HttpHeaders headers,
+      @RequestBody String payload) {
+    CloudEvent cloudEvent =
+        CloudEventHttpUtils.toReader(headers, () -> payload.getBytes()).toEvent();
+    try {
+      eventingService.processCloudEventRequest(cloudEvent);
```
tlawrie, line 83: if we make this non blocking as well i.e. CompletableFuture or similar, then I was reading up that we would just respond with a 202. In that the event is accepted but we are still potentially processing.

That appears to be a more common async loosely coupled event approach, than needing to respond to the event.

corcoja: @tlawrie, I would leave this method as blocking since `processCloudEventRequest()` does some validation and verification on the CloudEvent. If something's not looking right, an exception is raised by the service and is further returned as a response to the client invoking this API, providing extra information on what's wrong.

#### Couple of `Exception` examples

- `BoomerangException`
- `InvalidPropertiesFormatException`
- `CloudEventsError`
- Invalid Token
- Other...

tlawrie: Hey @corcoja for the internal controller that should be fine. Particularly because its for inbound HTTP based events.

The async was mainly a comment for those that we are creating whilst running a workflow, such as the status events.

tlawrie:
Hey @corcoja can we get the change log at the top of the PR split into Eventing changes vs General changes as there are 47 files touched but I think probably only 15 related to eventing.

This will make it easier to ensure the relevant peices are migrated to v4 (Engine Service)

FYI @gchickma @amhudson

AndreiOla:
Since the 2 split branches were Merged, I will close this PR.

PR #213, opened 2022-06-30
fix team return

amhudson:
Closes #

For users with platform role of `user`, PATCH /workflow/workflow/{workflowId}/properties returns a 403 due to the teams being set to `null`. 

#### Changelog

**New**

- N/A

**Changed**

- N/A

**Removed**

- Removing ` flowUser.setTeams(null);` from iUserIdentityServiceImpl.getCurrentUser()

#### Testing / Reviewing

Tested and verified locally. Tagged 3.4.64.


amhudson:
@morarucostel This will need to go to dev and be included with our 3.9.0 upgrade

tlawrie (approved):

PR #219, opened 2022-07-10
Fix slackextensionsigning

tlawrie:
Closes #https://github.com/boomerang-io/flow.service.workflow/issues/222


tlawrie:
Merging in after discussion with @gchickma and low blast radius of only affecting the extensions.

PR #231, opened 2022-08-26
fix: speed up /teams api performance when team list is in large size

shenqinb-star:
Closes #

We have upgraded lasted flow version for CICD, and found the /teams API is very slow for admin user who has about 300+ teams. The general idea of the changes is using database batch query instead of loop teams one-by-one database query, then construct the API response.

#### Changelog

**New**

- {{new thing}}

**Changed**

- {{change thing}}

**Removed**

- {{removed thing}}

#### Testing / Reviewing

I have tested it locally by connecting with cicd flow dev database, it works well.


shenqinb-star:
@gchickma @tlawrie @marcusdroy hi team, can you help to review this PR, it is for /teams API speed up when the team list is in large size

shenqinb-star:
@amhudson hi Adriennem, No, we don't  have a WorkFlowRevisionCount collection. the data model WorkFlowRevisionCount is used to store the result for  aggregation query of workflows_revisions.  you can refer to the following methods of FlowWorkflowVersionRepository:
```
  @Aggregation(pipeline = {
		  "{'$match':{'workFlowId': {$in: ?0}}}",
	      "{'$sort':{'workFlowId': -1, version: -1}}",
	      "{'$group': { _id: '$workFlowId', 'count': { $sum: 1 }}}"
	})
  List<WorkFlowRevisionCount> findWorkFlowVersionCounts(List<String> workflowIds);

  @Aggregation(pipeline = {
		  "{'$match':{'workFlowId': {$in: ?0}}}",
	      "{'$sort':{'workFlowId': -1, version: -1}}",
	      "{'$group': { _id: '$workFlowId', 'count': { $sum: 1 }, 'latestVersion': {$first: '$$ROOT'}}}"
	})
  List<WorkFlowRevisionCount> findWorkFlowVersionCountsAndLatestVersion(List<String> workflowIds);
```

src/main/java/io/boomerang/mongo/entity/WorkFlowRevisionCount.java
```diff
@@ -0,0 +1,46 @@
...
+import org.springframework.data.mongodb.core.mapping.Document;
+
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.annotation.JsonInclude.Include;
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+@JsonIgnoreProperties(ignoreUnknown = true)
+@JsonInclude(Include.NON_NULL)
+@Document(collection = "#{@mongoConfiguration.fullCollectionName('workflows_revisions')}")
```
amhudson, line 12: @shenqinb-star if this isn't intended to be a collection, you should remove 

`@Document(collection = "#{@mongoConfiguration.fullCollectionName('workflows_revisions')}")`

and move this out of `src/main/java/io/boomerang/mongo/entity/`

amhudson (changes requested):

src/main/java/io/boomerang/mongo/entity/WorkFlowRevisionCount.java
shenqinb-star: OK, make sense

shenqinb-star: @amhudson hi Adrienne, I have moved it to package `mongo/model`, see commit https://github.com/boomerang-io/flow.service.workflow/pull/231/commits/54e159f685838aca8a0a678199b98cbca8da319b

amhudson: If WorkFlowRevisionCount isn't related to an entity in `../mongo/entity`, it shouldn't be in `../mongo/model`. Perhaps `io/boomerang/model` is more appropriate.

shenqinb-star: Although WorkFlowRevisionCount is not related to a collection entity, but it is another kind of representatives for the result of aggregation query which I think it should be prat of mongo.
BTW What's the differences between `../mongo/entity` and `../mongo/model`? I saw some similar models in package `../mongo/model`,

amhudson: 1.  `.../mongo/entity/...` and  `.../mongo/model/...` establish the structure of the database records. `.../mongo/entity/...` contains the structure of the database collections while `.../mongo/model/...` are the classes used to support the collections. 

Although `WorkFlowRevisionCount` presents data based on `RevisionEntity`, it is not a mongo model or entity and should not be included in `.../mongo/...`. 

2. The queries defined in `.../mongo/service/...` and `.../mongo/repository/...` are ok as they return data based on `RevisionEntity`.

tlawrie: Hey @amhudson as this is _kind of_ a mongo aggregation of sorts, could we stick it in /mongo/model as its an abstraction on a mongo entity but also quite coupled to the mongo entity.

tlawrie: Also, @amhudson  great pick up on the annotations!

tlawrie:
Provisionally assigned to Flow 3.12.0

amhudson (approved):

tlawrie (approved):

PR #234, opened 2022-08-30
add a new internal API getTeamWorkflows

Amy0511:
Closes #

{{short description}}

#### Changelog

**New**

- {{new thing}}

**Changed**

- {{change thing}}

**Removed**

- {{removed thing}}

#### Testing / Reviewing

### DIFF
--- a/src/main/java/io/boomerang/controller/InternalController.java
+++ b/src/main/java/io/boomerang/controller/InternalController.java
@@ -122,5 +122,4 @@ public class InternalController {
       @RequestParam String propertyKey) {
     return workflowService.getRevisionProperties(workflowId, workflowVersion, taskId, propertyKey);
   }
-
 }
--- a/src/main/java/io/boomerang/controller/InternalController.java
+++ b/src/main/java/io/boomerang/controller/InternalController.java
@@ -122,4 +122,10 @@ public class InternalController {
       @RequestParam String propertyKey) {
     return workflowService.getRevisionProperties(workflowId, workflowVersion, taskId, propertyKey);
   }
+  
+  @GetMapping(value = "/workflows/{teamId}")
+  public List<WorkflowShortSummary> getTeamWorkflows(@PathVariable String teamId) {
+    return workflowService.getWorkflowsShortSummaryForTeam(teamId);
+  }
+
 }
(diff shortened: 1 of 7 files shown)


tlawrie:
Hi @Amy0511, we are actually trying to move away from internal APIs and deprecate them. Internal APIs were around from the days that we didn't have the `/apis` endpoints and only had the endpoints supporting the web application. Now that we have APIs, we should be able to satisfy any requirements and utilize either a Global or Team token to access.

FYI @gchickma

Amy0511:
@tlawrie I have changed the code to add new `api` in `WorkflowsController` . I didn't find the existing `api` to get workflows for a team, but there is an existing similar function. So just added `api` definition in controller. Please help to review it. Thanks.

amhudson (approved):

tlawrie:
Hi @Amy0511 the API controllers are actually under `io.boomerang.controller.api` so you'll probably need to add this level of command to the `TeamsV1Controller.java` with a context similar to `/teams/{teamId}/workflows` and you'll need to determine the scope that it should be protected at.

If you choose the global scope then you'll need a global API token. This makes sense for any integration from an end system.

If you choose the team scope then you'll need to do an additional check that the API token being passed in, has access to the teamId also being passed in.

tlawrie (changes requested):
See comments

Amy0511:
@tlawrie Hi Tyson, do you mean that we should NOT add APIs under `io.boomerang.controller`? 
The getTeamWorkflows API is called by CICD and CICD uses `UserAuthentication` (NOT `TokenBasedAuthentication`) to call Flow API.

tlawrie:
Hi @Amy0511 that makes sense, I assumed API meant Token Authenticated APIs.

In that case, move it under TeamController.java as typically an ID would follow the plural of the previous context i.e. `/workflow/{workflowId}` not `/workflow/{teamId}` and therefore better off putting it under `/teams/{teamId}/workflows`

Amy0511:
@tlawrie I have changed the code and moved the API to TeamController. Please help to reivew the code again. Thanks.

tlawrie (approved):

Amy0511:
image
@tlawrie I can't merge the PR. Could you please help on that?

PR #238, opened 2022-09-15
fix retrieve teams by HLG when externalUserUrl is configured

shenqinb-star:
Closes #

A defect was found on Core Flow with following configuration: non-admin user always get empty team list
```
flow.externalUrl.team=
flow.externalUrl.user=${users.base.url}/internal/users/user.
```
Based on above configurations and code logic, the team Id list is get from `flow.externalUrl.user` which is mapped to **_higherLevelGroupId_** in **_flow_team_** collection, but the code logic is query against **__id_**.

#### Changelog

**Changed**

- For the case `flow.externalUrl.user` is configured but `flow.externalUrl.team` is not configured.  query  **_flow_team_** collection against **_higherLevelGroupId_** instead of **__id_*.

### DIFF
--- a/src/main/java/io/boomerang/mongo/service/FlowTeamService.java
+++ b/src/main/java/io/boomerang/mongo/service/FlowTeamService.java
@@ -16,5 +16,7 @@ public interface FlowTeamService {
   TeamEntity findById(String id);
 
   List<TeamEntity> findActiveTeamsByIds(List<String> ids);
+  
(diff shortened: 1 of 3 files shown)


tlawrie (approved):

tlawrie:
Hey @amhudson this looked good to me. Can you add your thoughts?

amhudson:
Looks good to me also!

amhudson (approved):

Issue #127, opened 2021-11-08
Refactor Output Properties to allow more reliable saving


marcusdroy:
Instead of saving workflow output properties directly on the running workflow, save them on each individual task execution, then after workflow is completed, save them on the workflow activity, this should allow them to be saved more consistently and support parallelism.

tlawrie:
Hi @marcusdroy were you want to have this attached to a particular release or is it just for the backlog?

timrbula:
@amhudson Also to take a look at this along with other issue

PR #240, opened 2022-09-21
Output properties refactor

amhudson:
Closes #

https://github.com/boomerang-io/flow.service.workflow/issues/127
https://github.com/boomerang-io/roadmap/issues/362
https://github.com/boomerang-io/roadmap/issues/342

#### Changelog

**New**

- N/A

**Changed**

- Moving initial recording of output properties from ActivityEntity to TaskExecutionEntity

**Removed**

- N/A

#### Testing / Reviewing

Tested and verified

### DIFF
--- a/src/main/java/io/boomerang/service/refactor/TaskServiceImpl.java
+++ b/src/main/java/io/boomerang/service/refactor/TaskServiceImpl.java
@@ -416,16 +416,10 @@ public class TaskServiceImpl implements TaskService {
       activity.setOutputProperties(new LinkedList<>());
     }
 
-    ObjectMapper objectMapper = new ObjectMapper();
 
-    try {
-      System.out
-          .println("***saveWorkflowProperty*******" + objectMapper.writeValueAsString(activity.getOutputProperties()));
-    } catch (JsonProcessingException e) {
-
-    }
 
     List<KeyValuePair> outputProperties = activity.getOutputProperties();
+
     String input = task.getInputs().get("value");
     String output = task.getInputs().get("output");
     KeyValuePair outputProperty = new KeyValuePair();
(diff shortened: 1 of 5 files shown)


src/main/java/io/boomerang/service/refactor/TaskServiceImpl.java
```diff
@@ -549,6 +563,17 @@ private void finishWorkflow(ActivityEntity activity) {
...
 
+    List<TaskExecutionEntity> taskExecutions =
+        taskActivityService.findTaskActiivtyForActivity(activity.getId());
+    if (activity.getOutputProperties() == null) {
+      activity.setOutputProperties(new LinkedList<>());
+    }
+
+    for (TaskExecutionEntity taskExecution : taskExecutions) {
+      if (taskExecution.getOutputProperties() != null) {
+        activity.getOutputProperties().addAll(taskExecution.getOutputProperties());
```
tlawrie, line 574: Hey @amhudson would there be a possibility of incorrectly overriding the same parameter by chance? if the same key is used across multiple Set Result Parameter tasks...

tlawrie:
Hey @amhudson looks great and thanks for the explanation on Slack of 

> TaskServiceImpl.saveWorkflowProperty() has the updates to record to the task.  
> TaskServiceImpl.finishWorkflow() records the task output properties to the workkflow activity

Made one query above inline.

src/main/java/io/boomerang/service/refactor/TaskServiceImpl.java
tlawrie: Ok. So is that another problem. What happens when you try to reference or use them? 🤔

amhudson: I don't think we have any validations to prevent the same key(s) from being used

tlawrie: Alright, so what I am hearing is that its not going backwards from what we have in use of the workflow results. Its just not working how it should in the long term. Thats ok, we can create a backlog item for it.

tlawrie: https://github.com/boomerang-io/flow.service.workflow/issues/241

tlawrie (approved):

PR #245, opened 2022-11-09
fix updating multi password-type workflow parameters issue

shenqinb-star:
hotfix for bug [All workflow password-type parameters will be set to null in database after creating/updating a parameter](https://github.com/boomerang-io/roadmap/issues/369)


amhudson (approved):

tlawrie:
One thing, prior to merging this in, is that this touches an area that's been forked and refactored.

PR #246, opened 2022-11-11
use default value for password-type parameter if user input is null

shenqinb-star:
another PR of hotfix for bug [All workflow password-type parameters will be set to null in database after creating/updating a parameter](https://github.com/boomerang-io/roadmap/issues/369).

This PR fix password-type parameters bug when executing workflow


amhudson (approved):

gchickma (approved):
Looks good.  Thanks @shenqinb-star

PR #249, opened 2022-11-14
set default appName for external Nav API

shenqinb-star:
Set default value When external Navigationav API doesn't return `appName`,

### DIFF
--- a/src/main/java/io/boomerang/service/profile/LaunchpadNavigationServiceImpl.java
+++ b/src/main/java/io/boomerang/service/profile/LaunchpadNavigationServiceImpl.java
@@ -2,6 +2,8 @@ package io.boomerang.service.profile;
 
 import java.util.ArrayList;
 import java.util.List;
+
+import org.apache.logging.log4j.util.Strings;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.annotation.Qualifier;
 import org.springframework.beans.factory.annotation.Value;
@@ -122,6 +124,12 @@ public class LaunchpadNavigationServiceImpl implements LaunchpadNavigationServic
     HttpEntity<String> requestUpdate = new HttpEntity<>("", headers);
     ResponseEntity<NavigationResponse> response =
         restTemplate.exchange(uriComponents.toUriString(), HttpMethod.GET, requestUpdate, NavigationResponse.class);
+    NavigationResponse result = response.getBody();
+    if (result != null && result.getPlatform() != null && Strings.isBlank(result.getPlatform().getAppName())) {
+      // set default appName if the external Navigation API does NOT return appName.
+      String appName = settingsService.getConfiguration("customizations", "appName").getValue();
+      result.getPlatform().setAppName(appName);
+    }
     return response.getBody();
   }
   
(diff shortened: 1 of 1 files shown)


gchickma (approved):
thx @shenqinb-star

amhudson (approved):

src/main/java/io/boomerang/service/profile/LaunchpadNavigationServiceImpl.java
```diff
@@ -122,6 +126,19 @@ private NavigationResponse getExternalNavigationResponse(String email) {
...
+        // set default appName if the external Navigation API does NOT return appName.
+        String appName = settingsService.getConfiguration("customizations", "appName").getValue();
+        appName = Strings.isBlank(appName)? APP_DEFAULT_NAME: appName;
+        result.getPlatform().setAppName(appName);
+      }
+      if(!Strings.isBlank(result.getPlatform().getPlatformName())) {
+        // add | to the end of the platformName
+        result.getPlatform().setPlatformName(result.getPlatform().getPlatformName() + " |");
+      }
+    }
```
tlawrie, line 141: Hi @shenqinb-star @amhudson @gchickma should this just be defaulted in the Settings via the Loader?

I'm inclined to say that we should not be defaulting in the code.

tlawrie (changes requested):
Please default in the loader.

src/main/java/io/boomerang/service/profile/LaunchpadNavigationServiceImpl.java
shenqinb-star: @tlawrie, the source code logic is
1. if appName exists in the Settings, it will default to Settings
2. if appName does not exist in the Settings, it will default to "Boomerang Flow" in the code.
Do you think is this logic more strong?

tlawrie: I think my concern here is we have been working to remove any hard coded strings from the code that hold a name.

Am I missing a scenario where the default wouldn't be in Settings from the loader?

shenqinb-star: @tlawrie, currently, Flow for CICD doesn't have the appName in the settings.

shenqinb-star: if we default to Settings, we need to have loader to add AppName in settings for Core Flow and CICD Flow.

shenqinb-star: @tlawrie hi tyson, I have changed the code to default appName in settings

shenqinb-star:
@tlawrie hi Tyson, can you pls review again, I made the updates to default to Settings.

shenqinb-star:
@tlawrie hi Tyson, I have updated the code to make it default in Settings, can you review again?

tlawrie (approved):

amhudson:
@shenqinb-star is there a loader update "customization" setting?

> @tlawrie hi Tyson, can you pls review again, I made the updates to default to Settings.

PR #258, opened 2022-12-22
Feature/rebase eventing

AndreiOla:
**Changed**

Added eventing.enabled option for fixing an issue when NATS server is not used.


AndreiOla:
@corcoja  I disabled EventingService, when the corresponding property is set to false, but I am not sure that workflow will still work.
Is  "public void processCloudEventRequest(CloudEvent cloudEvent)" method still used by workflow when NATS server is not used/disabled? If so, do you think it would be a good idea to extract this functionality to a different service like: InternalEventingService and have just this method in it?

corcoja:
@AndreiOla, `public void processCloudEventRequest(CloudEvent cloudEvent)` is still used by InternalController, see `public ResponseEntity<EventResponse> acceptEvent(...)`.

AndreiOla:
@corcoja That is the point. `CloudEventsService` is not disabled when nats.eventing.enabled=false. Only the NATS part of event processing is enabled/disabled.

corcoja:
> @corcoja That is the point. `CloudEventsService` is not disabled when nats.eventing.enabled=false. Only the NATS part of event processing is enabled/disabled.

Ah, okay, I misunderstood the question.

I was aware of this limitation, since when I implemented Eventing with NATS, the possibility to turn off Eventing was not on the radar. It’s up to you to propose a solution for this, but keep in mind that disabling Eventing also means not using NATS for communication between `service.listener` and `service.workflow` and resorting to using APIs instead. And vice versa if enabling Eventing. So there is no need to create an Internal Eventing Service to be used when Eventing is off. When Eventing is disabled, it’s disabled across the entire application.

amhudson (approved):

AndreiOla:
I am closing this, since tag 3.4.95 contains all the changes made here and more.

PR #262, opened 2023-02-24
enhance for setting appName for external navigation API

shenqinb-star:
When external navigation API does not return `appName`, use `appName` in DB settings as default. Previous logic will throw exception when `appName` is not configured in DB settings, so a db loader is always required.

This PR makes `appName` as optional in DB settings which I think is more flexible.  When `appName` is not configured in DB settings, it will display as previous instead of throwing exception


shenqinb-star:
@tlawrie @gchickma @amhudson hi guys, I did an enhancement for setting default appName for external nav API, detail to see PR description. pls help to review.

amhudson (approved):

tlawrie:
Hey @amhudson do I need to somehow get this into v4 as well?

tlawrie:
Hey @amhudson do I need to somehow get this into v4 as well?

amhudson:
@tlawrie yep, think so

gchickma (approved):

PR #429, opened 2026-09-15
Submit validates the run timeout: default, floor at the task critical path, quota ceiling

tlawrie:
Decision 0063 requires a run timeout to be at least the transport timeout of the work it guards, and closes by noting that nothing validates the inequality at submit. This delivers that check, on every path that creates a run.

Before this, the workspace-facing submit clamped `timeout` to the workspace quota and nothing else looked at it. A run could therefore start with no timeout at all (the engine's `runworkflow` child submit applied no default when the revision carried none), or with a guard shorter than the budgets of its own tasks — in which case the watcher reaps healthy work at the run deadline.

## The three rules

`workflow/RunTimeoutPolicy.resolve` settles the value, in minutes:

| Rule | Behaviour |
| --- | --- |
| Default | The request's timeout wins, else the revision's, else the run-duration quota itself — the workspace's own `maxWorkflowRunDuration` override when it has one, else the platform `workspaces`/`max.workflowrun.duration` setting. A run is no longer created unguarded because nobody named a budget, and there is no second number for an operator to keep in step with the ceiling. |
| Floor | The value must be at least the revision's critical path — the longest chain of declared task timeouts through the graph. Below it the submit is refused, 400 `WORKFLOWRUN_TIMEOUT_TOO_SHORT` (1306), with a message naming the minutes required. Tasks with no declared timeout contribute 0; a timeout of 0 means unguarded and is exempt. |
| Ceiling | The owning workspace's `max.workflowrun.duration` quota clamps whatever clears the floor. A ceiling below the floor leaves no valid value, so it is refused with the same error naming the quota rather than silently clamped. |

No new setting and no change unit: the quota settings that already exist answer "how long may a run take here", per workspace and with a platform fallback, so default and ceiling come from one number and cannot drift. The `task` group's own `default.timeout`, which is the per-task budget, is untouched.

## Every submit path

The policy sits in `WorkflowService.submit(String workflowId, WorkflowSubmitRequest, boolean, String)` — the one method all five paths funnel through, so none can skip it:

| Path | Reaches it via |
| --- | --- |
| `POST /api/v2/workspace/{workspace}/workflow/{name}/submit` | `WorkspaceWorkflowControllerV2:314` → `submit(team, name, …)` → `internalSubmit` |
| Schedules | `ScheduleJob:82` → `submit(team, name, …)` → `internalSubmit` |
| Webhooks and events | `WebhookEventService:116,137,185` → `submit` / `internalSubmitForTeam` → `internalSubmit` |
| Slack action | `SlackService:377` → `submit(team, name, …)` → `internalSubmit` |
| A `runworkflow` child | `TaskExecutionService:824` → the chokepoint directly |

Two consequences of moving the ceiling out of `internalSubmit` and into the chokepoint:

- It is now resolved from the Workflow's owning workspace rather than the team segment the caller arrived on, so a `runworkflow` child gets the same ceiling as everyone else instead of none.
- The `HAS_WORKFLOW` edge records the workspace's ref while the quota lookup is by name, so the resolution goes through `getSlugByRefForType`. Without that the ceiling quietly fell back to the platform default for every submit.

A Workflow no workspace owns has no quota to apply, so it carries no ceiling — only the floor and the platform default constrain its runs.

## Retry

A retry creates a new run, and it was the one way past the workspace's quotas — and the one run whose timeout could outlive a ceiling the operator had since lowered. `WorkflowRunService.retry` now runs the same gate before it writes anything, so a refusal leaves the source run untouched:

- The concurrent-run, monthly-run and workspace-size check a submit runs. `WorkflowService.canRunWithQuotas` becomes the public `assertRunQuotas`, called by both — no second copy of the rule.
- The clone's `timeout` is clamped to the current run-duration ceiling.

The floor is deliberately not re-checked: a retry re-runs the same revision with the same request, so its critical path cannot have moved.

The engine's automatic retry runs inside the timeout path, where the source run is already terminal. A refusal there logs at warn (`[id] Retry refused: …`) and leaves the run terminal rather than failing the sweep that got there; a user-initiated retry surfaces the 400 `QUOTA_EXCEEDED` as any submit does. `WorkflowRunService` takes `WorkflowService` `@Lazy` — the same cycle break it already makes for `TaskExecutionService` — and resolves the owning workspace's name from the ref the ownership edge records, since the quota lookups are by name.

## Verification

Targeted, on Java 25; the full suite is the orchestrator's to run.

| Suite | Result |
| --- | --- |
| `SubmitRunTimeoutPolicyTest` (new, 6 tests) | the default applies and is the workspace's own quota rather than the platform value every workspace starts from; a timeout below the critical path is rejected with the code and a message naming 50 minutes; a timeout at the critical path passes; the quota still clamps; a quota below the floor is rejected naming the quota; `internalSubmit` takes the same path |
| `RetryRunQuotaTest` (new, 3 tests) | a retry at the concurrent-run limit is refused with `QUOTA_EXCEEDED` and creates no clone; a retry after the ceiling was lowered gets the lowered timeout; an auto-retry refused by quota leaves the run `timedout`, creates no clone and does not throw |
| `WorkflowRunRetryOwnerTest`, `WorkflowRunTimeoutGuardTest`, `StandaloneQuotaEnforcementTest`, `StandaloneQuotaConcurrentSubmitTest`, `EngineModeWorkflowSubmitTest` | 25 tests, green |
| `WorkflowRunLifecycleTest`, `WorkflowWatcherTest`, `ClaimTimeoutReapTest`, `TriggerAgnosticWorkspacesTest`, `ScheduleLineageSubmitTest`, `WorkflowRunCancelActionsTest`, `QueuedTaskCancelCascadeTest` | 30 tests, green |
| `WebhookEventAuthorizationTest`, `WebhookEventEmptyDataTest`, `ScheduleJobTest`, `ScheduleWatcherTest`, `WorkflowWorkspaceSpecDefaultsTest`, `WorkflowTaskParamValidationTest` | 29 tests, green |
| `LoaderMigrationTest` (6), `V3DumpMigrationTest` (1) | green |

Two fixtures gained a seed line rather than free-riding on whatever another class had already written into the shared Testcontainers Mongo: `WorkflowRunRetryOwnerTest` now seeds the quota settings and the `workspaceQuotas` feature toggle its retries read, and it already needed the quota settings because the engine's direct submit path now reads the ceiling every other path already read.

`execution-model.md` records the three rules in Timeouts and the quota behaviour in Retry. `api-contract.md` lists code ranges rather than individual codes, so it needs no change. Decision 0063 is immutable and its consequence is now delivered, so no new decision.

## One thing to flag

The shipped `task` / `default.timeout` is 90 minutes while `workspaces` / `max.workflowrun.duration` is 30, so out of the box every task's effective budget is three times its run's guard — 0063's inequality, violated by the seed itself. The floor here counts only per-task declared timeouts, so this does not trip it. Worth a separate look at the seeded values.


tlawrie:
Full `service-core` suite after the default-from-quota and retry-quota rework: Tests run: 439, Failures: 0, Errors: 0, Skipped: 0 (BUILD SUCCESS).

PR #431, opened 2026-09-15
A docker executor runs tasks on one host, for the no-Kubernetes quickstart

tlawrie:
Adds a third `TaskExecutor` — `dispatcher.executor=docker` — that runs task containers on one Docker
host, so Boomerang Flow runs end to end with `docker compose` and no Kubernetes. Decision
`specifications/decisions/0076-a-docker-executor-runs-tasks-on-one-host-for-the-no-kubernetes-quickstart.md`.

## What each executor supports

| | `tekton` (default) | `kube-jobs` | `docker` |
| --- | --- | --- | --- |
| Runtime object | Tekton v1 `TaskRun` | `batch/v1` `Job` | One container on the host daemon |
| Needs a cluster | yes, plus Tekton Pipelines | yes | no |
| Params | `PARAM_<NAME>` env vars | same | same (same `KubeHelperService.createTaskEnvVars`) |
| `RESULTS_PATH` | `/tekton/results` (directory) | `/dev/termination-log` | `/results.json`, copied out of the exited container |
| Timeout | `spec.timeout` | `activeDeadlineSeconds` | enforced by the executor — it stops the container and reports `DeadlineExceeded` |
| Status reasons | `DeadlineExceeded`, `JobDeleted`, `JobFailed`, `OOMKilled`, `ImagePull` | same | same |
| Script task | inline step script | ConfigMap at `/scripts/script` | tar copied in before start, same `/scripts/script` |
| Workspaces | PVC at `/workspace/<type>` | PVC at `/workspace/<type>` | named volume at `/workspace/<type>`, same labels |
| `/data` scratch | `emptyDir` | `emptyDir` | the container's writable layer (tmpfs when in-memory is asked for) |
| Runtime class / node selector / tolerations / host aliases / pull secret | yes | yes | none — no equivalent on a Docker host |
| Workspace size / class / access mode | applied | applied | recorded on the run, ignored (a local volume has none) |
| Resource limits | none | none | `dispatcher.docker.memory` / `dispatcher.docker.cpus` when set, else none |
| Cancel | overwrite the status condition | delete the Job | stop, then remove the container |

`kube.task.deletion` (`Never`/`OnSuccess`/`Always`) is read by `TaskService` and applies unchanged — the
Docker executor's `delete` removes the container by label.

## Workspace storage behind one interface

`WorkspaceReconciler` listed `PersistentVolumeClaim`s and read their labels itself, so releasing storage only
worked against a cluster. It now goes through `WorkspaceStore` — `create`, `exists`, `delete`, and `held()`
returning ref/type pairs (`service-dispatcher/src/main/java/io/boomerang/dispatcher/WorkspaceStore.java`).
`KubeService extends` it; `DockerWorkspaceStore` backs a workspace with a named volume
(`bmrg-flow-vol-ws-<type>-<ref>`) carrying exactly the labels the claim carries —
`boomerang.io/tier=workspace`, `boomerang.io/workspace-ref`, `boomerang.io/workspace-type`,
`boomerang.io/product` — so the same release path finds and deletes either. Provisioning is unchanged: a run
that declares workspaces still waits for the dispatcher to provision, then `start` (decision 0067).

Task logs got the same treatment: `LogService` held a `KubeLogService` and now holds a `TaskLogStore`, with a
Docker implementation that follows a running container's log and reads a finished one once.

`KubeServiceImpl` and `KubeLogService` are now conditional on the executor not being `docker`, and take the
Kubernetes client lazily, so the dispatcher starts with no cluster and no kubeconfig.

## Local stack

`docker-compose.docker.yml` layers `service-dispatcher` onto the main stack with
`DISPATCHER_EXECUTOR=docker` and `/var/run/docker.sock` mounted:

```
docker compose -f docker-compose.yml -f docker-compose.docker.yml up -d --build service-dispatcher
```

The dispatcher starts task containers as **siblings** on the host daemon. This is not Docker-in-Docker: task
containers show up in `docker ps` beside the stack's own, and workspace volumes in `docker volume ls`.

**Security.** Mounting `/var/run/docker.sock` gives that container root-equivalent control of the host daemon
— it can start any container, with any mount, as root. That is acceptable for a laptop quickstart and is **not
a production posture**; a real deployment runs a Kubernetes executor, where a task is a Pod under the
cluster's own RBAC. The overlay also switches dispatcher authentication off on `service-core`, the same
documented dev/test escape hatch `docker-compose.kube.yml` uses. Both facts are stated in the overlay's header
and in `specifications/architecture.md`.

## Dependency

`com.github.docker-java:docker-java-core` and `docker-java-transport-httpclient5`, pinned to **3.7.1** (the
latest stable) — neither is managed by the Spring Boot BOM, so the version is a property in
`service-dispatcher/pom.xml`. The aggregate `docker-java` artifact was not used: it drags in the Jersey and
Netty transports as well. `RestConfig`'s proxy and trust settings are for HTTP to core and are untouched; the
Docker socket does not go through them.

## Verification

- dispatcher suite 91/91 (`mvn -pl service-dispatcher test`)
- real-Docker test **run** — `DockerExecutorOnDockerHostTest` (3/3) executed real `busybox:1.36` containers on
  this machine's daemon, asserting a result read back from `RESULTS_PATH`, a log containing the task's output,
  a non-zero exit reported as `JobFailed` with its code, and a workspace volume created, mounted, found by
  `held()` and deleted. It is `@EnabledIf` on a reachable socket and skips cleanly where none answers.

## Not included

- No e2e Playwright scenario for the docker executor (the kube one is `e2e/tests/dispatcher-kube.spec.ts`).
  The live JUnit test covers the executor contract; a compose-level journey is a follow-up.
- Azure Container Apps, explicitly deferred — noted in decision 0076.
- No CI job runs the live Docker test; it skips wherever no daemon answers.


tlawrie:
Up to date with `feat-v5` (two merge commits), and one integration the merge exposed is fixed here.

**What conflicted and how it resolved.** `KubeHelperService` — the correctness fixes added Kubernetes label-value sanitisation and stopped a user label overwriting a `boomerang.io/*` selector, while this branch had moved workspace listing behind `WorkspaceStore`; both kept, and the Docker path inherits the sanitisation because it labels its volumes through the same helper. `KubeJobsExecutor` — kept the unparseable-termination-message handling that fails an oversize result with `ResultsTooLarge` alongside this branch's executor changes. The specification files and the decisions index carry both sides.

**The integration the merge exposed.** `DockerExecutor` read its image, command and script straight off the TaskRun spec. An `ai` task deliberately carries none of them — the dispatcher supplies the worker image for that type — so an `ai` task dispatched to a Docker deployment would have failed with no image. It now goes through `TaskImageResolver`, exactly as the two Kubernetes executors do.

Result reading takes the parser's `Optional`, with semantics that differ from Kubernetes on purpose: there is no 4096-byte ceiling on the Docker path (the file is read whole off the container filesystem), so an unparseable payload is never a truncation symptom — it means the task wrote something that is not a results payload, which is logged with no results rather than failed. An oversize but well-formed payload stays the engine's to reject at `end`.

Verification: `service-dispatcher` 111 tests, 0 failures, 0 errors (up from 80 as the merge brought in the AI-type and label-sanitisation tests).

tlawrie:
Merged `feat-v5` (982b0b4) so this branch carries #438's `TaskResourceResolver`, then two changes.

## One property set sizes every runtime (a8c997c)

This branch predated #438 and had grown its own sizing knobs, so after the merge an operator would have
sized Kubernetes tasks and Docker tasks with two different property names for one intent. Removed:

| Gone | Was |
| --- | --- |
| `dispatcher.docker.memory`, `dispatcher.docker.cpus` | `application.properties:68-69` |
| their two `@Value` fields | `DockerExecutor.java:86-89` |
| `DockerExecutor.bytes(String)`, a local Docker-CLI size parser | `DockerExecutor.java:444` |
| `dockerCliSizesBecomeByteCounts` | `DockerExecutorTest` |

`DockerExecutor` now takes `TaskResourceResolver` in its constructor beside `TaskImageResolver` and calls
`memoryLimitBytes()` / `cpuLimitNanos()` where the host config is built, so `kube.resource.limit.*` is the
only sizing set and `16Gi` / `500m` mean the same thing on all three executors. Blank still means "no limit
set", the resolver's own definition — Docker then imposes none, exactly as blank sets no Kubernetes limit.
Docker ignores the `ephemeral-storage` pair (no counterpart on a Docker host) and the requests (no scheduler
to inform).

`grep -rn 'dispatcher.docker.memory|dispatcher.docker.cpus'` over the tree: **zero hits**.
`docker-compose.docker.yml` never set either, so it needed no change. Updated `service-dispatcher/README.md`
and the container-resources section of `specifications/task-runtime.md` to name one property set for all
three executors.

Two new tests replace the parser test — `kube.resource.limit.*` set becomes the container's memory and
NanoCPUs, blank sets neither — and the `DockerExecutorTest` / `DockerExecutorOnDockerHostTest` constructors
are updated. `mvn -pl service-dispatcher -am test`: **121 tests, 0 failures** (`DockerExecutorTest` 15,
`DockerExecutorOnDockerHostTest` 3 against the real daemon, `DockerExecutorSelectionTest` 2,
`DockerWorkspaceStoreTest` 5).

## A Playwright scenario for the Docker executor (e98bc0f)

The no-Kubernetes quickstart is the first path a new user takes and had unit tests plus one real-Docker
JUnit test only — nothing proved a run end to end on it. `e2e/tests/dispatcher-docker.spec.ts` runs against
`docker-compose.yml` + `docker-compose.docker.yml`, gated on `E2E_DOCKER_DISPATCHER=true` the way the
Kubernetes scenarios gate on `E2E_DISPATCHER`:

1. a catalogue task runs in a container, the run reaches `succeeded`, and its log streams back;
2. a declared param arrives in the container as `PARAM_<NAME>`;
3. a result written to `RESULTS_PATH` comes back on the TaskRun (on Docker that is a copy-out from the
   exited container, a path the Kubernetes runtimes never exercise);
4. a `workflowRun` workspace is a Docker volume labelled `boomerang.io/workspace-ref=<runId>`, seen while
   the run is live and gone within one reconcile interval after it completes;
5. a task exiting non-zero ends the run `failed` with `statusReason=JobFailed`, from the closed set of
   decision 0065.

It reuses `e2e/support/dispatcher.ts` rather than copying it; only the two runtime-specific assertions are
local, shelling out to `docker volume ls --filter label=…` the way the Kubernetes spec shells out to
`kubectl`. `docker-compose.docker.yml` already sets `FLOW_DISPATCHER_WORKSPACE_RECONCILE_MS: "10000"`, so
the release assertion needed nothing added. Invocation documented in `CLAUDE.md` beside the Kubernetes one
and in the local-stack paragraph of `specifications/architecture.md`.

### It has not been executed yet

I built the jars and the images and brought the overlay stack up, and it could not start: three of the four
host ports the stack publishes are held by other long-running work on this machine.

```
Error response from daemon: failed to set up container networking: driver failed programming external
connectivity on endpoint flow-idpzero-1: Bind for 127.0.0.1:4380 failed: port is already allocated
```

`:4380` and `:7700` belong to a separate `cheer-*` compose stack that has been up for six days, and `:3000`
to a VS Code plugin host. None of them are mine to take. Remapping is not a workaround here: the OIDC
issuer is pinned to `http://idp.localhost:4380` in `docker/mongo/seed-auth-oidc.js` and the redirect URI to
`localhost:3000` byte-for-byte in `docker/idpzero/server.yaml`, so the sign-in that `auth.setup.ts` performs
requires those exact ports.

Everything short of execution passes: `npx playwright test tests/dispatcher-docker.spec.ts --list`
enumerates all five scenarios, and the `docker volume ls --filter label=…` invocation the spec shells out to
was verified against this daemon. The stack was torn down afterwards and the other stack left untouched.
**Please run it once on a host with those ports free before merging** — that is the one claim in this PR I
cannot make myself.

PR #432, opened 2026-09-15
Thin status CloudEvents by default, and sink authentication by header

tlawrie:
Two changes to the outbound CloudEvent sink: status events stop carrying the whole run, and a sink can authenticate with a request header instead of a URL secret.

## 1. Thin status events by default

`io.boomerang.event.status.workflowrun` / `.taskrun` put the full public run model in `data`: every parameter value and every result (4 KB each, decision 0041) went to every URL in `flow.events.sink.urls`. A consumer that uses the event as a trigger and reads the run back got nothing from that payload and inherited its leak surface.

`flow.events.sink.payload` now selects the shape, defaulting to `thin` — the run's identity and lifecycle, nothing else:

```json
{ "specversion": "1.0", "type": "io.boomerang.event.status.taskrun",
  "source": "/apis/v1/events", "subject": "/taskrun/68b1.../status/failed",
  "id": "9f1c...", "time": "2026-09-16T04:21:07Z", "datacontenttype": "application/json",
  "data": { "id": "68b1...", "workflowRef": "66aa...", "workflowRunRef": "67cc...",
            "status": "failed", "statusReason": "JobFailed", "phase": "completed",
            "labels": { "initiatorId": "user-1" },
            "creationDate": 1758000000000, "startTime": 1758000001000, "duration": 4210 } }
```

No `params`, no `results`, no `annotations`. `payload=full` restores the previous model. Both shapes are built by one projection step, `RunStatusSummary.project`, called from `EventFactory` for both event types; the CloudEvent `type`, `source`, `subject` and `id`, and the outbox rows behind them, are unchanged, so routing and filtering built on the envelope are unaffected.

A workflow run has no `statusReason` field (it is a task-end concept, decision 0065) and no `workflowRunRef`, so its thin payload omits both; the models carry no `endTime`, so `startTime` + `duration` express it.

Also removed: `WorkflowRunStatusEvent.toCloudEvent` logged the entire serialised payload at INFO on every delivery.

## 2. Sink authentication, additive

`flow.events.sink.urls` was the only way to configure a sink, so the receiver's secret had to ride in the query string, where every forward proxy and access log records it. That form is unchanged and still delivered to, URL untouched — some receivers can only be given a URL.

A sink can now be declared as a destination with a header instead:

| Property | Meaning |
| --- | --- |
| `flow.events.sink.enabled` | Unchanged; egress is off by default |
| `flow.events.sink.payload` | `thin` (default) or `full` |
| `flow.events.sink.urls` | Unchanged: comma-separated bare URLs |
| `flow.events.sink.destinations[n].url` | A sink that authenticates with a header |
| `flow.events.sink.destinations[n].header-name` | Defaults to `Authorization` |
| `flow.events.sink.destinations[n].header-value` | The secret; the `application.properties` comment says to supply it from the environment |

The header value is sent only on the request to its own sink and is never logged. Both lists are delivered to, through the same configured `internalRestTemplate`, so `RestConfig`'s proxy routing and per-template timeouts still apply (decision 0062).

### Property binding

`EventSinkService`'s two `@Value` fields become one bound record, `EventSinkProperties` (`flow.events.sink.*`) — the first `@ConfigurationProperties` in `service-core`, with `@ConfigurationPropertiesScan` added to `Application`, as boomerang-io/flow#325 (b) asks for. That issue's other targets (`flow.watcher.*`, `flow.events.outbox.*`, `flow.schedule.*`) are untouched here: `OutboxDispatcher` still reads its interval and its `@ConditionalOnProperty` from the raw keys.

## Specs

- `api-contract.md` — the webhook section splits into "Events in" (unchanged) and a new "Events out": the envelope, both `data` shapes with the JSON example, and the two sink forms with the reason to prefer a header.
- `execution-model.md`, `architecture.md` — payload default noted; the `application.properties` line citation moved.
- Decision `0079-status-events-carry-identity-and-lifecycle-not-the-run-model.md` and its `decisions/README.md` row.

Nothing in `client-web/` consumes these events — the UI polls the API, and its only CloudEvents references are the workflow editor's inbound-trigger help text (`Configure.tsx:445-483`, `ConfigureEventTrigger.tsx:56`).

## Verification

`mvn -pl service-core -am test -Dtest='io.boomerang.event.**'` — 17 tests, 0 failures:

| Test | Tests | Covers |
| --- | --- | --- |
| `StatusEventPayloadTest` (new) | 5 | Both payload shapes serialised and asserted by key set: thin workflow-run and task-run key sets exactly, thin never carrying `params`/`results`/`annotations` or any of their values, full still the whole model, and the same envelope either way |
| `EventSinkDeliveryTest` (new) | 5 | A destination sends its header and leaves the URL alone, a custom header name is honoured, a bare `?token=` URL sink sends no `Authorization` and is POSTed verbatim, both lists are delivered to while a disabled sink sends nothing, and delivery is wired to `internalRestTemplate` |
| `OutboxDeliveryTest` (existing) | 2 | Unchanged: one row per status change, none for a phase-only claim, delivered and marked sent. Boots the full context, so it also proves the new bound record binds there |
| `WebhookEventAuthorizationTest`, `WebhookEventEmptyDataTest`, `GenericStatusEventTest` (existing) | 5 | Ingress and the generic event payload, unaffected |

Full suite: run by the orchestrator.


tlawrie:
Full `service-core` suite on this branch: 446 run, 0 failures, 0 errors, 0 skipped (BUILD SUCCESS).

PR #433, opened 2026-09-16
Eight correctness fixes from live feat-v5 use

tlawrie:
Eight correctness items reported by an integrator running `feat-v5` live, each verified against the code before being fixed. All eight were live on `feat-v5`; none was already fixed.

| # | What was wrong | Fix | Test |
| --- | --- | --- | --- |
| 1 | A workflow node other than `start`/`end` with no `taskRef` dereferenced null in `WorkflowService.convertTaskSlugsToRefs` and answered 500. | New `WORKFLOW_MISSING_TASK_REF` (1211, 400, "Task "{0}" has no task reference."). The reference is required at save — `create`, `apply` and `composeApply` all funnel through the same conversion — and re-checked against the stored revision at submit, because a workflow saved earlier can still carry one. Both name the node. | `WorkflowTaskRefValidationTest` — create, blank ref, apply, and submit of a stored revision whose node lost its ref |
| 2 | Run labels were passed to Kubernetes unchecked, so a value containing `/` made Kubernetes reject the whole object and the run's PVC failed with a 422 the author never saw. | `KubeHelperService` sanitises every user-supplied key and value in the one place all executors share (`getTaskLabels`, `getWorkflowLabels`, `getWorkspaceLabels`). Values and key names: ≤63 of `[A-Za-z0-9._-]`, alphanumeric ends, other characters → `_`. Key prefixes: DNS subdomain, ≤253 of `[a-z0-9.-]`, other characters → `-`. A key with no usable name is dropped; every alteration is logged at debug. The `boomerang.io/*` and `app.kubernetes.io/*` labels the dispatcher sets are placed first and a user label never replaces one — they are the selectors every lookup, watch and delete runs on. | `KubeHelperServiceTest` — the reported `/` case, a >63-char value, key sanitising and end-trimming, the no-overwrite rule, an unusable key, and the dispatcher's own labels unchanged |
| 3 | No `DELETE /api/v2/task/{name}` on the global catalogue (405); only the workspace-scoped route existed. | `TaskControllerV2.delete` → new `TaskService.deleteGlobal`, same shape as the workspace route: resolve the name through the caller's own relationship, then the shared `delete`, which refuses with `TASK_DELETE_IN_USE` (409) while a run in flight references the Task. `@AuthCriteria(TASK, DELETE, {global, user, session})`. | `TaskCatalogueDeleteTest` — delete removes it, an in-flight run refuses it, an unknown name is rejected |
| 4 | In engine mode `GET /api/v2/workspace/system` was 404 even though the loader seeds that workspace. | `WorkspaceService` is standalone-only because it composes members, quotas and insights (`InsightsService` is itself standalone-gated), so un-gating it would drag the whole workspace module into engine mode. Added `EngineWorkspaceService` + `EngineWorkspaceControllerV2` (engine-mode only) reading `GET /{workspace}` and `/query` straight off the workspaces collection, password params filtered. Writes stay standalone-only. `EngineWorkspaceInterceptorConfiguration` now also guards `/api/v2/workspace/{workspace}`, so any name other than `system` is refused there too (`/query` binds a handler with no `{workspace}` variable and passes through). | `EngineWorkspaceReadTest` — service resolves `system`, 404s `other`, query returns only `system`; MockMvc serves `/workspace/system` (200) and refuses `/workspace/other` (404). Bean presence/absence pinned in `FlowModeGatingEngineModeTest` and `FlowModeGatingTest` |
| 5 | Every "does not exist" answer was a 400 under a `*_INVALID_REF*` code, because one enum constant served both "no such resource" and "blank/malformed input". | The family answers `404` and means only the first; blank/malformed guards move to the matching `*_INVALID_REQ` code on 400. Four new: `TEAM_INVALID_REQ` (1104), `WORKFLOW_INVALID_REQ` (1212), `TASK_INVALID_REQ` (1406), `PARAMS_INVALID_REQ` (1803); `WORKFLOWRUN_INVALID_REQ`, `TASKRUN_INVALID_REQ`, `SCHEDULE_INVALID_REQ` already existed. A reference that fails to resolve *inside a body* stays 400. Decision `0080` + README row, `api-contract.md` error section. Classification below. | `EngineWorkspaceInterceptorTest` and `WorkspaceWorkflowUniquenessTest` updated to the new status/reason; `EngineWorkspaceReadTest` asserts 404 on the code path |
| 6 | A new Task version looked invisible to existing workflows. | Confirmed deliberate and documented, no behaviour change. `taskVersion` is stamped at save in `WorkflowService.createWorkflowRevisionEntity` (`TaskService.retrieveAndValidateTask` resolves the node's own version, or latest when it asked for none); `DAGUtility.createTaskList` resolves the same way at run time, so a stored null still resolves latest, and records the result on the TaskRun. Written into `task-runtime.md` with how to move a workflow forward. | — (documentation) |
| 7 | Results over the cap "succeeded with every result nulled and no signal". | Root cause is the `kube-jobs` executor, not the engine. Kubernetes truncates a termination message at 4096 bytes; the truncated prefix is broken JSON; `TerminationMessageParser` returned an empty list for it, so the dispatcher reported "no results" and the engine's own `flow.engine.task.results.max-bytes` check (which does work, `TaskRunService.end`) never saw a payload. The too-large probe only ran when the message was *blank*, which truncation never produces. The parser now returns `Optional.empty()` for an unparseable message, and `KubeJobsExecutor.readResults` runs the too-large probe unconditionally and fails the task with `ResultsTooLarge` when the probe fires or the unparseable message is at the ceiling. A short unparseable message is a task writing something that is not a results payload: no results, logged, not failed. Tekton already failed the TaskRun itself on overflow — unchanged. | `KubeJobsExecutorTest` — a completed Job whose pod carries a 5000-byte truncated message fails with `ResultsTooLarge`; parser tests separate "unparseable" from "no results" |
| 8 | Dead outbox rows were revivable only by editing Mongo. | `GET /api/v2/system/outbox?status=&page=&limit=` lists rows (dead by default) and `PUT /api/v2/system/outbox/replay?ids=&status=&olderThan=` resets them to `pending` with `retry.after=now` and `attempts=0`, answering `{"replayed": n}`. Both `@AuthCriteria(SYSTEM, …, {global})`. Independent of `flow.events.sink.enabled` so an operator can see and requeue rows a disabled sink left behind. `api-contract.md` and `execution-model.md` updated. | `OutboxDeliveryTest.aDeadRowIsReplayedAndThenDelivered` — dead → listed → replayed → delivered on the next drain |

**Review gate:** not triggered. Nothing here touches `DAGUtility` or `TaskExecutionService`.

**Data model:** none. No new field, index, collection or migration. Item 8 reads and updates the existing `events_outbox` documents only.

## Item 5 — classification of every `*_INVALID_REF*` throw site

Rule applied (decision `0080`): **404** when the unresolved reference is the one the request path addresses; **400** when the input is blank/malformed, or when the reference is inside a request body or query.

Enum constants moved to `NOT_FOUND`: `TEAM_INVALID_REF`, `WORKFLOW_INVALID_REF`, `WORKFLOWRUN_INVALID_REF`, `TASK_INVALID_REF`, `TASK_INVALID_WORKFLOWREF`, `TASKRUN_INVALID_REF`, `ACTION_INVALID_REF`, `SCHEDULE_INVALID_REF`, `PARAMS_INVALID_REFERENCE`, `INTEGRATION_INVALID_REF`. `WORKFLOW_INVALID_TASK_REF` deliberately stays 400 — it is a reference inside a saved workflow body.

Throw sites moved to a 400 code (all were blank/null/malformed guards unless noted):

| File | Lines | From → to |
| --- | --- | --- |
| `workflow/WorkflowService.java` | 253, 382, 398*, 550, 573, 712, 737, 815, 831, 848, 1374, 1571, 1830, 1925, 1958 | `WORKFLOW_INVALID_REF` → `WORKFLOW_INVALID_REQ` (*398 is the duplicate-name conflict on create, not a missing reference) |
| `workflow/WorkflowService.java` | 1285, 1360 | `TASK_INVALID_REF` → `WORKFLOW_INVALID_TASK_REF` with the node and the value — a `runworkflow` node's `workflowRef` that does not resolve is a body reference, so it keeps 400 |
| `workflow/WorkflowRunService.java` | 481, 804, 843, 859, 875, 896, 923, 1025 | `WORKFLOWRUN_INVALID_REF` → `WORKFLOWRUN_INVALID_REQ` |
| `workflow/WorkflowRunService.java` | 351 | `TASKRUN_INVALID_REF` → `TASKRUN_INVALID_REQ` |
| `workflow/TaskService.java` | 595, 620, 714 | `TASK_INVALID_REF` → `TASK_INVALID_REQ` |
| `workspace/WorkspaceService.java` | 152, 255, 266, 332, 340, 502, 540, 608, 648, 735, 760 | `TEAM_INVALID_REF` → `TEAM_INVALID_REQ` |
| `schedule/ScheduleService.java` | 196, 409 | `SCHEDULE_INVALID_REF` → `SCHEDULE_INVALID_REQ` (null request, or a body `workflowRef` that does not resolve) |
| `workflow/ParameterService.java` | 99, and `update`/`delete` restructured | `PARAMS_INVALID_REFERENCE` → `PARAMS_INVALID_REQ` for the null/nameless request |
| `integrations/GitHubService.java` | 113, 126 | `ACTION_INVALID_REF` → `INTEGRATION_INVALID_REF` — an installation lookup failing was reported as an Action error, and `INTEGRATION_INVALID_REF` had no throw site at all |

Every other site in the family stays on its `*_INVALID_REF*` code and now answers 404 — the "path names something that does not exist, or the caller cannot reach it" cases in `WorkflowService` (284, 586, 725, 762, 1389, 1836, 1947, 1961), `WorkflowRunService` (195, 408, 429, 458, 493, 837, 853, 869, 886, 917, 987, 1031), `TaskRunService` (747, 783, 852, 889), `WorkflowExecutionService` (66, 106, 133), `TaskService` (150, 171, 656, 738, 901), `WorkflowTemplateService` (56, 62), `ActionService` (100, 114, 254, 265), `ScheduleService` (89, 102, 172, 314, 600), `WorkspaceService` (168, 270, 274, 346, 506, 510, 544, 548, 612, 616, 629, 739, 743, 764, 768), `WebhookEventService` (131) and `EngineWorkspaceInterceptor` (44).

Three defects surfaced while classifying and are fixed in the same commit:

- `ParameterService.delete` threw `PARAMS_INVALID_REFERENCE` unconditionally *after* deleting, so deleting a global parameter always errored even on success.
- `ACTION_INVALID_REF` and `SCHEDULE_INVALID_RUNONCE` had no matching key in `messages.properties` (the file carried `ACTION_INVALID_REFERENCE` and `SCHEDULE_INVALID_RUNONE`), so both rendered as `No message available`.
- `GitHubService` reported installation failures under an Action code.

**Frontend:** no change needed. `client-web` has no `status === 400` check anywhere; every API call goes through axios (`Config/serverFetch.ts`, `Config/servicesConfig.ts`), which rejects 400 and 404 identically, and the MSW mocks already answer 404 for a missing resource.

## Left open

- **Item 6 — "follow latest" is not built.** Pinning at save is deliberate (reproducibility) and is now documented. A per-node "always use latest" option is the open ask. The editor does expose the upgrade: `TemplateNode` renders a "New version available" prompt from the `upgradesAvailable` flag and `TaskUpdateModal` stamps the new version on save — an explicit per-node upgrade, not a free version picker.
- **Item 8 — no last-error field.** `EventOutboxEntity` stores no failure text (and no `deadAt`); the delivery error only reaches the log. The list route returns id, refType, ref, status, attempts and `occurredAt`. Adding an error field would be a data-model change, so it is not in this PR.
- **Item 3 — neither delete path removes the relationship node.** `TaskService.delete` deletes the Task and its revisions but leaves the `TASK`/`TEAMTASK` relationship node behind; the new global route matches the workspace route exactly, as directed. Worth a separate issue.
- `TASK_INVALID_NAME` (1403) is used for "no such task, or no access" in `TaskService`'s scoped lookups and stays 400; it is outside the `*_INVALID_REF*` family decision 0080 covers.
- `WorkflowRunService.query` still throws when a workspace resolves to no workflows instead of answering an empty page (left on 404 `WORKFLOWRUN_INVALID_REF`, status unchanged from the caller's point of view except 400→404).
- `OutboxControllerV2` returns `EventOutboxEntity` rather than a separate wire model, and the outbox routes live in the `event` package rather than on `SystemControllerV2` — `core` must not import a feature package.


tlawrie:
Full `service-core` suite on this branch after the test fix in `e477a0d69`: Tests run: 451, Failures: 0, Errors: 0, Skipped: 0 (BUILD SUCCESS).

PR #434, opened 2026-09-16
feat: a first-class ai task type executed as a pod from a Boomerang-published worker image

tlawrie:
A first-class `ai` task type: an author declares it and never builds a container; the dispatcher resolves a Boomerang-published worker image (`boomerangio/task-ai`) and runs it as a pod (or Docker container) like any other dispatched task. Decision `0078` records why this shape (declared type + platform-resolved image, the Kubeflow pattern) rather than an in-dispatcher model call (n8n/Langflow/Dify): the prompt is untrusted content and must not share a process with the dispatcher's credential, and a dispatcher deployment registered for `ai` alone is the egress zone.

## What the author sees

A catalogue task `ai` (category "AI") with the standard surface — `endpoint` (OpenAI-compatible base URL), `token` (password), `model`, `systemPrompt`, `prompt`, `temperature` (new `slider` input, 0–2, step 0.1, default 0.7), `maxTokens`, `responseFormat` (`text`|`json`), `seed`, `files` (workspace paths read into context) and `maxContextBytes` — and six results: `output`, `promptTokens`, `completionTokens`, `totalTokens`, `finishReason`, `model`. Usage is metered as typed results on the TaskRun; a platform sums `totalTokens` over the run query. `output` is subject to the 4 KB result cap like any result.

## Halves

| Module | Change |
| --- | --- |
| `lib-common` | `TaskType.ai`; `ConfigType.slider`; nullable `min`/`max`/`step` on `AbstractParam` (the slider's range; absent on every other param) |
| `service-core` | `ai` joins the dispatched types. **Review gate:** `TaskExecutionService` touched in two places — the `execute` switch arm (`case template, script, custom, generic, ai`) and the `queue` auto-start guard, so an `ai` task waits for a dispatcher instead of auto-executing; `DAGUtility` read, unchanged. `WorkflowWatcher.REQUEUEABLE_TYPES` gains `ai` (agent-executed, so a killed claimant is recoverable). |
| `service-loader` | Seeded task + revision (catalogue 88 tasks / 131 revisions); `_0047__SeedAiTask` for existing installs (`_0022` reconciles only on first run). Verified against the real v3 dump. |
| `service-dispatcher` | `TaskImageResolver` (`executor/`): for type `ai`, image = `flow.dispatcher.ai.image` (default `boomerangio/task-ai:latest`) and command `prompt`, ignoring anything on the spec; used by both Kubernetes executors and the `NO_TASK_IMAGE` gate. The default dispatcher now registers `template,custom,script,ai` so a single-dispatcher install runs AI tasks; the AI network zone is the opt-in (drop `ai` from the general deployment, run a second registered for `ai` only — `service-dispatcher/README.md`). |
| Worker image (not in this repository) | The `ai` worker image lives in `boomerang-io/tasks` as `tasks/ai` and publishes `boomerangio/task-ai` — `boomerangio/task-ai:latest` and `boomerangio/task-ai:<version>`, tagged from `@boomerang-io/task-ai@<version>` — like every other task image, on its own version line rather than the product tag. wolfi-base + bash/curl/jq with one `prompt` command: it builds the chat-completions call from `PARAM_*` env, appends `files` under a byte budget, retries 429/5xx, fails hard on 4xx without printing the token, and writes the six results to either results channel (Tekton directory or termination-log JSON). The param and result contract is shared with the seeded catalogue task here, so the two repositories move together. This branch adds no image to `ci-release.yml`: the product tag still builds the four product images (decision `0006`). |
| `client-web` | `slider` param input (existing `Components/Slider`, now fraction-capable) wired through DynamicFormik's custom-component hook and the Task Manager's param-type form; `ai` node type, "AI" palette category with the `MachineLearningModel` icon; config modal, results table and Task Manager list needed nothing bespoke. tsc 11 → 9. |
| Specs | `task-runtime.md` (AI tasks section, param/result tables, image resolution, the zone pattern, the env-var note), `execution-model.md`, `data-model.md`, `design-system.md`; decision `0078` (titled "a Boomerang-published worker image" — it is unmerged, so it is edited as a draft rather than superseded). |

## Data model

`TaskType.ai`, `ConfigType.slider`, three nullable range fields on the param spec, one seeded task/revision, one change unit. No entity field, no index.

## Known limitation, stated in the docs

`token` is a `password` param: scrubbed on API reads, but it reaches the pod as a plain environment variable until per-task secrets exist. Found while verifying that scrub: `WorkflowRunService.filterSensitiveValues` joins against the *workflow revision's* params, so a password value typed literally into a node param whose `password` type is declared only on the catalogue task is not scrubbed — pre-existing, covers 84 catalogue `password` params, documented in `task-runtime.md` and decision 0078; the test pins the supported pattern (declare the secret as a workflow param and reference it). Raised separately.

## Verification

service-core 437 run / 0 failures / 0 skipped; service-dispatcher 80/80; service-loader 8/8 including the real v3 dump; the worker image's own shell test and end-to-end run are verified in `boomerang-io/tasks`, not here; client-web vitest 398 passed (2 pre-existing timing flakes in `Schedules.spec`/`WorkspaceDetailed.spec`), build exit 0, tsc 9.


tlawrie:
**Merge-order dependency: the worker image does not exist yet.**

`flow.dispatcher.ai.image` defaults to `boomerangio/task-ai:latest`, and that repository has no tags today (`boomerangio/task-flow:latest` resolves; `boomerangio/task-ai:latest` returns `object not found`). The image is built from `boomerang-io/tasks`, where the worker is still an open PR.

What has to happen before an `ai` task can pull an image: the worker lands on `boomerang-io/tasks` `main`, and a `task-ai@1.0.0` tag builds and pushes `boomerangio/task-ai:1.0.0` and `:latest`. Flow can then pin the default to the exact version instead of `:latest`.

This does not block merging to `feat-v5`, which is unreleased: an `ai` task submitted before the image exists fails its pod on image pull and surfaces as `statusReason=ImagePull`. It does block shipping the `ai` task type in a release, so it belongs on the 5.0.0 checklist.

The param and result contract is already mutual and needs no change on either side. Flow's dispatcher emits one `PARAM_<NAME>` per param plus the `PARAM_NAMES` manifest of original names, and sets `RESULTS_PATH` (`service-dispatcher/.../KubeHelperService.java:114-146`); `@boomerang-io/task-core` on that repository's `main` reads `PARAM_NAMES` to map `PARAM_PRIVATEKEY` back to `privateKey` exactly, and honours `RESULTS_PATH`, branching on directory (Tekton) versus single JSON file (plain Jobs, Docker). The seeded catalogue task in this PR is unaffected.

*Corrects an earlier version of this comment, which described a publication order requiring `@boomerang-io/task-core` to be released to npm first, and the older `@boomerang-io/task-ai@1.0.0` tag format. Neither is current: task images build from the pnpm workspace, and the tag patterns are now `<name>@<version>`.*

Issue #439, opened 2026-09-16
Param substitution silently nulls any value containing a quote or newline


tlawrie:
## What happens

A param whose value contains a double quote or a newline resolves to `null`. No error reaches the
run — the task starts with the param empty and fails on its own required-param check, which sends
you looking in the wrong place entirely.

## Why

`ParameterManager.replaceStringInObject` (`service-core/.../engine/ParameterManager.java:348-364`)
JSON-encodes the value, substitutes **raw text** into that encoded string, then parses it back:

```java
String objectString = objectMapper.writeValueAsString(object);
final StringSubstitutor substitutor = new StringSubstitutor(replacements, "$(", ")");
String replacedObjectString = substitutor.replace(objectString);
return objectMapper.readValue(replacedObjectString, Object.class);
} catch (Exception e) {
  LOGGER.error(e.toString());     // logged
}
return null;                      // and the param is silently empty
```

Substituting into an already-encoded JSON string means the replacement text must itself be
JSON-escaped. It is not, so a `"` closes the string early and a newline is an illegal literal in
one — `readValue` throws, the catch logs and drops it, and the caller
(`resolveParamLayers`, line 105: `p.setValue(resolveParam(...))`) stores `null`.

## Reproducing

Submit a WorkflowRun with a multi-line param and reference it from a node:

```jsonc
// submit
{"params":[{"name":"instruction","value":"Line one.\nReturn JSON: {\"ok\": true}"}]}
// node
{"name":"prompt","value":"$(workflow.params.instruction)"}
```

The TaskRun's `prompt` comes through as `null`. The same reference with a single-line,
quote-free value (`https://openrouter.ai/api/v1`) resolves correctly, which is what makes the
failure so confusing to chase.

## Why it matters now

This is reachable from the feature shipped in 5.0.0-beta.4. The `ai` task type's `systemPrompt`
and `prompt` are `texteditor::text` — multi-line prose is their whole purpose, and any prompt that
asks for JSON output contains quotes. So **an `ai` node cannot be driven by workflow params at
all**: every realistic prompt hits this. Found while wiring Cheer's first `ai` node; worked around
by inlining prompts into the workflow definition at apply time, which avoids substitution
entirely.

## Suggested fix

Substitute before encoding, not after — resolve against the value itself and let Jackson encode
the result once. If the current shape is worth keeping, escape each replacement as a JSON string
body before handing it to `StringSubstitutor`.

[...]

PR #441, opened 2026-09-22
fix(engine): param substitution no longer nulls quoted or multi-line values, and an object param keeps its structure

tlawrie:
Param substitution no longer nulls a value that contains a quote or a newline — a multi-line prompt, a JSON body, a shell script or a task result with a trailing newline now reaches the task byte for byte.

A second commit fixes a defect of the same family found while reviewing the first: an `object`-typed param whose value is a *structure* containing references collapsed to one leaf's value. See "Second defect" at the end.

Fixes #439

## The mechanism

`ParameterManager.replaceStringInObject` did three things in sequence:

1. `objectMapper.writeValueAsString(object)` — JSON-encoded the value, so a string became a quoted JSON literal with its own quotes and newlines escaped.
2. `StringSubstitutor` replaced each `$(...)` token with the replacement's **raw** `toString()`, spliced straight into that encoded text with no JSON escaping.
3. `objectMapper.readValue(replacedObjectString, Object.class)` — re-parsed.

A replacement carrying a `"` or a newline made step 3's input malformed JSON, the `catch (Exception e)` logged and fell through to `return null`, and `resolveParam` took that null as the resolved value.

## Before / after

Workflow param `prompt`:

```
Summarise the text below.
"The quick brown fox jumps over the lazy dog."
Answer as JSON: {"summary": "..."}
```

referenced from a `custom` node's `shellScript` as `$(params.prompt)`:

| | TaskRun `spec.script` the dispatcher receives |
| --- | --- |
| Before | `null` — the container gets no script at all |
| After | `#!/bin/sh` / `cat <<'EOF'` / the prompt verbatim / `EOF` |

## The fix

Substitute into string leaves directly, where quoting and newlines are not special: a `String` is run through the substitutor as-is; a `Map` is walked (values, and string keys, preserving the whole-document substitution this replaced); a `Collection` or array is walked element by element; anything else (number, boolean, null) is returned untouched. The walk is depth-guarded at 32 so a pathological structure cannot exhaust the stack. No JSON encode, no re-parse.

The silent null is gone too: an unexpected failure returns the original value unchanged — verbatim passthrough, the same fallback an unmatched `tasks.<name>.results.<result>.<jsonpath>` already takes — and logs at error with the value's **shape** (`string[214 chars]`, `object[3 entries]`), never its content, because a password-typed param resolves through the same path. The exception's own message is not logged either: `StringSubstitutor`'s cycle guard quotes the value in it.

## Judgement call: a non-String replacement interpolated into a larger string

**Chosen: JSON-serialise it** (`{"k":"v"}`, `["a","b"]`) rather than keep Java's `toString()` (`{k=v}`, `[a, b]`).

- There is no established behaviour to preserve. Today `{k=v}` is spliced into the JSON-encoded text and usually fails the re-parse, so the param resolves to null; only a map of quote-free scalars survived, and then only as an unparseable `{k=v}`.
- JSON is already the convention for a structured param value on this path: the dispatcher JSON-encodes a non-string param on its way into `PARAM_<NAME>` (`specifications/task-runtime.md`, "What the container receives"), and an `object`-typed param's structure is JSON on the wire. `{k=v}` is a Java debug format nothing downstream can read.
- It is the only rendering a task can act on: a body interpolated into a `curl` argument or an AI prompt is usable as JSON and useless as `{k=v}`.

The structural path is untouched: a token that is the **entire** value of an `object`-typed param still returns the structure itself from `resolveParam`, before this method runs — pinned by `objectTypedParamKeepsTheStructureOfItsReplacement`. That short-circuit turned out to be too wide, which is the second defect below.

No decision record. This is a bug fix that restores the contract decision 0040 already states, and the rendering choice follows the JSON convention that decision and the dispatcher already set; a record would only restate them.

**Data model:** none.

**Review gate:** not triggered — `ParameterManager` only; `DAGUtility` and `TaskExecutionService` are untouched.

## Verification of the first fix

Written red first: on `origin/feat-v5` the new unit tests fail 13 of 28, 11 of them returning `null`; the end-to-end test fails with `spec.script` = `null`.

New in `ParameterManagerTest` (28 tests, all green):

`resolvesValueContainingADoubleQuote`, `resolvesValueContainingADoubleQuoteInsideALargerString`, `resolvesMultiLineValue`, `resolvesValueWithQuotesNewlinesAndABackslash`, `resolvesValueWithWindowsLineEndings`, `resolvesValueContainingALiteralDollar`, `resolvesValueContainingADollarParenSequence`, `selfReferencingValueFallsBackToTheOriginalValue`, `resolvesJsonDocumentValue`, `resolvesMultiLinePromptWithEmbeddedQuotes`, `resolvesMultipleReferencesInOneValue`, `resolvesTaskResultValueWithATrailingNewline`, `rendersAStructuredReplacementAsJsonWhenInterpolatedIntoAString`, `rendersAListReplacementAsJsonWhenInterpolatedIntoAString`, `objectTypedParamKeepsTheStructureOfItsReplacement`, `resolvesQuotedMultiLineValueIntoSpecFields`.

The existing characterization tests stay green unchanged, covering `params.<name>`, `params.<name>.<jsonpath>`, `<scope>.params.<name>`, `<scope>.params.<name>.<jsonpath>`, `tasks.<name>.results.<result>[.<jsonpath>]`, case-insensitive matching, spec-field substitution and unmatched-token passthrough.

New in `WorkflowRunLifecycleTest`: `aQuotedMultiLineParamReachesTheTaskAsItsLiteralValue` — submits a run carrying the quoted multi-line prompt, waits for the `custom` TaskRun to be admitted (`ready`/`pending`, the state a dispatcher claims), and asserts the literal value in both `spec.script` and the resolved `shellScript` param.

```
ParameterManagerTest, ParamRedactionTest, WorkflowRunLifecycleTest   45 run, 0 failures
ChildWorkflowRunTest, PayloadCapTest, AiTaskTypeTest,
TaskExecutionServiceRunScheduledWorkflowTest,
KeyedResultsAndEventArmingTest                                       24 run, 0 failures
```

`specifications/task-runtime.md` now states the substitution semantics under "What the container receives".

## Second defect: an `object` param whose value is a structure loses the structure

Same family as #439 — structural loss, no error — found while reviewing the first fix, and fixed in the second commit.

`resolveParam` returned on the **first** reference it resolved for any `object`-typed param:

```java
if (!Objects.isNull(foundValue)) {
  if (ParamType.object.equals(type)) {
    return foundValue;              // the FIRST matched reference's value
  } else {
    foundKeyValues.put(foundKey, foundValue);
  }
}
```

So an `object` param whose value is a structure containing references resolved to one leaf's replacement and nothing else:

| | `object` param `{"url": "$(params.host)/api", "retries": 3}`, `host` = `https://example.com` |
| --- | --- |
| Before | `"https://example.com"` — a String. The `/api` suffix gone, the `retries` key gone, the object gone |
| After | `{"url": "https://example.com/api", "retries": 3}` |

The discovery matcher runs over the Map's `toString()` (`{url=$(params.host), retries=3}`), finds `params.host`, and the early return threw the rest away.

**What the short-circuit is still for.** One case, and it must keep working: the param's value is a single reference and nothing else — stored as the String `"$(params.config)"` on a param declared `type: object`, where `config` resolves to a Map. There the point is to get the Map back rather than a stringified one. So the condition is narrowed to exactly that: the value is a `String` whose whole **trimmed** content is one match of the same `REGEX_DOT_NOTATION` the discovery loop uses — `matcher.find() && matcher.start() == 2 && matcher.end() == trimmed.length() - 1 && !matcher.find()`. One compiled `Pattern` constant now serves both, so the test and the discovery cannot drift apart; no second hand-rolled pattern. `$(a)$(b)` and `$(a)x` both fail the span check and fall through, as they should.

**Whitespace:** `"  $(params.config)  "` is treated as a single reference — the value is trimmed before the span check. A structure cannot carry leading or trailing spaces, so trimming loses nothing, whereas counting them as surrounding text would silently turn the param into a JSON string, which is the very change this guard exists to prevent.

**Everything else falls through** to the accumulation and `replaceStringInObject` — which is what finally activates the `Map`/`Collection`/array recursion the first commit introduced. Until now nothing reached those branches, because the only values that were structures were `object` params and they all returned early. String leaves and string keys are substituted in place; numbers and booleans are untouched; a leaf that is itself a whole reference to an object renders as JSON, the same rule the first commit chose for a structured replacement inside a string — one convention, not two.

**Discovery still matches over `originalValue.toString()`.** That stays, and it is safe: a reference in a nested value still appears in the Map's flattened text, so it is still found. The `toString()` text is only used to discover *which* references to resolve; substitution then walks the real object, and that is what decides the resulting shape. Changing discovery is not needed for this fix and is not in scope.

### Verification of the second fix

No existing test pinned the first-reference-wins behaviour, so nothing had to be changed to make the fix pass. The only test touching the short-circuit, `objectTypedParamKeepsTheStructureOfItsReplacement`, pins the case that is deliberately kept and stays green unmodified (its comment is reworded to say *why* it is the one case).

Seven new tests in `ParameterManagerTest`: `objectParamKeepsEveryKeyWhenAReferenceHasSurroundingText`, `objectTypedParamKeepsTheStructureDespiteSurroundingWhitespace`, `objectParamResolvesAReferenceInEveryLeaf`, `objectParamResolvesAReferenceNestedTwoLevelsDeep` (a Map inside a List inside a Map), `objectParamResolvesAReferenceInAMapKey`, `objectParamRendersAStructuredLeafReplacementAsJson`, `objectParamResolvesAReferenceInsideAList`.

```
ParameterManagerTest                                                 35 run, 0 failures
ParameterManagerTest, ParamRedactionTest, WorkflowRunLifecycleTest,
ChildWorkflowRunTest, AiTaskTypeTest                                 59 run, 0 failures
```

`specifications/task-runtime.md` now states the `object` rule beside the string-leaf rule. No decision record — the same reasoning as above applies.


tlawrie:
Full `service-core` suite on this branch: 488 run, 0 failures, 0 errors, 0 skipped (BUILD SUCCESS).

tlawrie:
Full `service-core` suite with both defects fixed: 495 run, 0 failures, 0 errors, 0 skipped (BUILD SUCCESS).

Issue #405, opened 2026-09-01
Run-submission storage quota check is a no-op: reflection on a Map spec always throws and is swallowed


tlawrie:
`WorkflowService.canRunWithQuotas` (`service-core/src/main/java/io/boomerang/workflow/WorkflowService.java:~935-989`) reads the workspace size by reflection: `ws.getSpec().getClass().getDeclaredField("size")` followed by `Integer.valueOf(size)`. At submission time `ws.getSpec()` is a `LinkedHashMap` (JSON-sourced), so `getDeclaredField("size")` throws `NoSuchFieldException` (the field is declared on `HashMap`, not `LinkedHashMap`) and the catch block swallows it — the run-submission storage quota check is a no-op today. Even when it did run, `Integer.valueOf("1Gi")` would throw.

Expected: convert the spec with the `ObjectMapper` and compare through `StorageQuantityUtil.toGi(...)` (added in commit `f076dc11f` for `setUpWorkspaceDefaults`), the same way the save-time check now does. Found while running the dispatcher-on-Kubernetes scenarios.

PR #443, opened 2026-09-23
fix: enforce the storage quota on run submission

tlawrie:
The run-submission storage quota refused nothing. It now reads the requested size the same way the save-time check does, and a submit whose workspace asks for more storage than the workspace's quota allows is refused with `QUOTA_EXCEEDED`.

Fixes #405

## Why it never fired

Two independent reasons, either one enough on its own:

1. `WorkflowWorkspace.spec` is declared `Object` and arrives as a `LinkedHashMap`. `getDeclaredField("size")` on the runtime class does not find `HashMap.size`, throws `NoSuchFieldException`, and the `catch (NoSuchFieldException | IllegalAccessException ex) { // Do nothing }` swallowed it. A `WorkflowWorkspaceSpec` instance fared no better - the field is private, so the read threw `IllegalAccessException` into the same catch.
2. Had it reached the comparison, `Integer.valueOf("1Gi")` throws `NumberFormatException` on any real Kubernetes quantity - and that is not in the catch, so a working reflective read would have turned the quota into a 500 rather than a refusal.

## Before / after

```java
// before - WorkflowService.canRunWithQuotas, once for "workflow" and again for "workflowrun"
try {
  Field sizeField = ws.getSpec().getClass().getDeclaredField("size");
  String size = (String) sizeField.get(ws.getSpec());
  if (Integer.valueOf(size) > quotas.getMaxWorkflowStorage()) { throw ... }
} catch (NoSuchFieldException | IllegalAccessException ex) {
  // Do nothing
}
```

```java
// after - one method, the rule resolveWorkspaceSpec already applies at save
private void checkWorkspaceStorageQuota(WorkflowWorkspace ws, CurrentQuotas quotas) {
  Integer maxSizeGi =
      switch (ws.getType()) {
        case "workflow" -> quotas.getMaxWorkflowStorage();
        case "workflowrun" -> quotas.getMaxWorkflowRunStorage();
        case null, default -> null;
      };
  if (maxSizeGi == null || ws.getSpec() == null) {
    return;
  }
  String size = this.objectMapper.convertValue(ws.getSpec(), WorkflowWorkspaceSpec.class).getSize();
  if (size != null && StorageQuantityUtil.toGi(size) > maxSizeGi) {
    throw new BoomerangException(
        BoomerangError.QUOTA_EXCEEDED, "Requested Workspace size", size, maxSizeGi);
  }
}
```

The conversion is the one `TaskExecutionService.getTaskWorkspaces` already uses on the same field; the Gi parse and the `QUOTA_EXCEEDED` arguments are the ones `resolveWorkspaceSpec` already uses at save. No third way was invented.

**The two blocks were not identical.** They differed in the type they matched and the quota they read - `workflow` against `max.workflow.storage`, `workflowrun` against `max.workflowrun.storage` - and were otherwise the same code. So both types were covered by the old code (neither worked), and both stay covered: the collapsed method picks the limit by type. That matches the save-time check, which sets up the same two types against the same two settings, so `workflowrun` being subject to its own quota is deliberate, not a newly opened gate.

A spec naming no `size` does not throw, matching the save path, where an unset size is filled in with the quota default and so can never breach it.

## Verification

`WorkflowSubmitStorageQuotaTest` (new, 5 tests): a `workflow` workspace over quota is refused and the exception names the size; one within quota is admitted; `3000Mi` is refused and `1500Mi` admitted against the same 2Gi quota, which only a real Gi conversion gets right; a spec with no size is admitted; a `workflowrun` workspace is held to `max.workflowrun.storage` while clearing `max.workflow.storage`.

- Before the fix: 5 run, 3 failed - the three over-quota submits all succeeded.
- After: 5 run, 0 failed.
- Neighbours, unchanged: `StandaloneQuotaEnforcementTest` (3), `WorkflowWorkspaceSpecDefaultsTest` (3), `StandaloneQuotaConcurrentSubmitTest` (2) - 8 run, 0 failed.

Specs: `performance.md` listed the storage quota as enforced at run submit only, which was the one place it was not; it and `task-runtime.md` now say save and submit, with the line anchors moved.

Data model: none.

Review gate: not triggered - no change to `DAGUtility` or `TaskExecutionService` (`getTaskWorkspaces` was read, not modified).


tlawrie:
Full `service-core` suite on this branch: 505 run, 0 failures, 0 errors, 0 skipped (BUILD SUCCESS).

Issue #404, opened 2026-09-01
Engine overwrites a TaskRun's per-task workspaces with all run workspaces after start


tlawrie:
`TaskExecutionService.getTaskWorkspaces()` (`service-core/src/main/java/io/boomerang/engine/TaskExecutionService.java:609-625`) overwrites `TaskRunEntity.workspaces` with **all** of the run's workspaces (ignoring the task's own declared subset) from inside `execute()`, which runs `@Async` after `start()` has already returned to the dispatcher's synchronous `PUT /api/v1/dispatcher/taskrun/{id}/start`. The dispatcher builds the pod from its claim-time snapshot, so this write never affects the pod — but it silently replaces the per-task list that `DAGUtility.createTaskList` now sets (commit `f076dc11f`), so the persisted TaskRun reports the wrong workspaces afterwards.

Expected: either delete the method (the materialised list is authoritative) or make it respect the task's own declared workspaces. Found while running the dispatcher-on-Kubernetes scenarios (`e2e/tests/dispatcher-kube.spec.ts`).

PR #444, opened 2026-09-23
fix(engine): a task keeps only the workspaces it declares

tlawrie:
The engine no longer replaces a TaskRun's declared workspaces with the run's whole set: `getTaskWorkspaces` is deleted and the materialised list from `DAGUtility.createTaskList` is the single source of truth.

Fixes #404

## Shape chosen: (a) delete the method

`WorkflowTask.workspaces` is already `List<TaskWorkspace>` (`lib-common/.../model/WorkflowTask.java:48`) — the exact type `TaskRunEntity.workspaces` holds (`lib-common/.../entity/TaskRunEntity.java:48`). `DAGUtility.createTaskList` copies it verbatim, `name`/`type`/`optional`/`mountPath` included (`service-core/.../engine/DAGUtility.java:213-214`, commit `f076dc11f`), so the deleted method did no work `createTaskList` does not already do.

The one thing it did differently — resolving `mountPath` from the **workflow-level** `WorkflowWorkspaceSpec` — was already inert on both paths:

| Path | Why the write never landed |
| --- | --- |
| The pod | The dispatcher snapshots the TaskRun when it claims it (`DispatcherService.getTaskQueue:243-249`); `execute()` runs `@Async` **after** `start()` has answered `PUT .../taskrun/{id}/start` |
| The database | `execute()` saves only inside `if (endTask)` (`TaskExecutionService.java:396-400`), and `endTask` is `false` for all five dispatched types — the mutated entity was simply discarded |

And nothing downstream needs that resolution: both executors default a blank `mountPath` to `/workspace/<type>` (`service-dispatcher/.../KubeJobsExecutor.java:307-310`, `.../TektonServiceImpl.java:230-234`).

Option (b) — intersecting the run's workspaces with the task's declared names — would rebuild, from a second source, a list that is already correct at the point it is created. Less code and one source of truth wins.

Two things the deletion also removes: a `NullPointerException` inside the async executor when a workflow's workspace carries no `spec` (`OBJECT_MAPPER.convertValue(null, …)` returns `null`, then `spec.getMountPath()`), which is every workflow created through the engine-level `create` that bypasses `setUpWorkspaceDefaults`; and a per-execution rebuild of a list nobody read.

## Review gate

Gated class: `service-core/src/main/java/io/boomerang/engine/TaskExecutionService.java`. Exactly two edits, both confined to the method and its single call site named in the issue. `DAGUtility` is untouched.

| Change | Semantics before | Semantics after |
| --- | --- | --- |
| Deleted `private void getTaskWorkspaces(TaskRunEntity, WorkflowRunEntity)` (was `:618-634`) | Rebuilt `taskExecution.workspaces` from every `wfRunEntity.getWorkspaces()` entry, taking `mountPath` from each one's `WorkflowWorkspaceSpec` | Gone. No caller remains (`grep getTaskWorkspaces` is empty) |
| Removed the call from `execute()`'s `case template, script, custom, generic, ai ->` arm (`:340-343`) | Arm ran the rebuild, then fell through with `endTask = false` | Arm is a comment only; still falls through with `endTask = false`. Every other arm, the entry compare-and-set, the `if (endTask)` save and `end()` are unchanged |

No import changed — the file imports `io.boomerang.common.model.*`.

## Data model: none

No field, index, collection or migration. `TaskRunEntity.workspaces` keeps its type and meaning; the fix is about who writes it.

## Verification

The regression tripwire is `service-core/src/test/java/io/boomerang/engine/TaskWorkspaceScopeTest.java`: a run declaring two workspaces (`workflow`, `workflowrun`) whose `echo` node declares only `workflowrun`, asserted at all three points — the materialised TaskRun, the dispatcher claim payload from `getTaskQueue` (the pod is built from that, not from the record), and the persisted TaskRun held for 3 s after `execute()` has moved it to `running`.

It passes on `feat-v5` as it stands, because the corrupting write is currently discarded rather than saved. To prove it actually catches the defect rather than passing vacuously, a `taskRunRepository.save(taskExecution)` was added after the `getTaskWorkspaces` call and the test was run against the unpatched method:

```
execute() must not replace the TaskRun's workspaces with the run's
  but held [workflow, workflowrun] ==> expected: <1> but was: <2>
```

That experiment was reverted; the method is deleted instead. The test now guards the invariant against both the persisted-write regression and any future widening on the claim path.

Targeted runs, all green — 44 tests, 0 failures, 0 errors:

`TaskWorkspaceScopeTest` (1), `WorkflowRunLifecycleTest` (9), `AiTaskTypeTest` (1), `DispatcherQueueClaimTest` (7), `DispatcherWorkspaceReleaseTest` (8), `EngineWorkspaceReadTest` (5), `StaleSnapshotSaveTest` (5), `WorkflowRunWorkspaceQueryTest` (4), `WorkflowWorkspaceSpecDefaultsTest` (3), `TriggerAgnosticWorkspacesTest` (1) — the last six being everything under `service-core/src/test` that names `setWorkspaces`/`getWorkspaces` on a run or task.

## Specification

`specifications/task-runtime.md` §Volumes and workspaces: the per-task list is now described as set once and never widened, and the paragraph records where `mountPath` actually comes from — the task's own declaration, with a `/workspace/<type>` fallback — noting that the workflow-level `spec.mountPath` is stored but read by nothing. `execution-model.md` needed no change: it describes workspace provisioning and teardown, not the per-task list. No decision record — this restores the behaviour decision `0067` and commit `f076dc11f` already assume.


tlawrie:
Full `service-core` suite on this branch: 501 run, 0 failures, 0 errors, 0 skipped (BUILD SUCCESS).

